PR #42, opened 2024-05-06
31 migrate from mongodb to mysqlmariadb

Metawolve:
I migrated the whole project, including the CLI toolbox, from the mongoDB driver to Entity Framework with MySQL/MariaDB.

During this migration, the repository pattern was removed, as Entity Framework already implements the repository pattern with the DBSets.

I also fixed up existing unit tests to work with Entity Framework.

This meant pretty substantial changes to the project, including changing some of the entities properties to actual relations, as we are now working with a relational SQL database instead of No-SQL.

I tested the core functionality of the API (creating, updating, querying and deleting different records) through Swagger and the CLI toolbox and fixed the issues I found.

There are some things I could not test properly, like the Telegram integration for example. As the changes in this issue are already pretty extensive, and as we need Entity Framework to work on some of the other open issues, I suggest that we merge these changes into main and go on from there by fixing issues we find and implementing further performance optimizations for Entity Framework in separate issues. A lot of the aspects I did not fully test yet are planned to be reworked or scrapped, like the CLI toolbox for example.


src/Eurofurence.App.Domain.Model/Abstractions/IEntityRepository.cs
```diff
@@ -1,50 +1,36 @@
...
-
-    public interface IEventConferenceTrackRepository : IEntityRepository<EventConferenceTrackRecord>
-    {
-    }
-
-    public interface IEntityStorageInfoRepository : IEntityRepository<EntityStorageInfoRecord>
-    {
-        Task<EntityStorageInfoRecord> FindOneAsync(string entityType);
-    }
+    //public interface IEntityRepository<TEntity> where TEntity : EntityBase
```
Rain336, line 11: I guess this file can also be deleted, since it's completely commented out

src/Eurofurence.App.Infrastructure.EntityFramework/AppDbContext.cs
```diff
@@ -0,0 +1,82 @@
...
+        public virtual DbSet<RoleRecord> Roles { get; set; }
+        public virtual DbSet<TopicRecord> Topics { get; set; }
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            base.OnModelCreating(modelBuilder);
+
+            modelBuilder.Entity<RegSysAlternativePinRecord>().Property(x => x.PinConsumptionDatesUtc)
+                .HasConversion(new ValueConverter<List<DateTime>, string>(
+                    v => JsonConvert.SerializeObject(v),
```
Rain336, line 74: Can we maybe add either [`Pomelo.EntityFrameworkCore.MySql.Json.Microsoft`](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql.Json.Microsoft/) or [`Pomelo.EntityFrameworkCore.MySql.Json.Newtonsoft`](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql.Json.Newtonsoft/) to get native json support with MySQL's json type?
All you would have to do is add one of them, register in `Startup` and call `.HasDataType("json")`

src/Eurofurence.App.Server.Services/DependencyResolution/AutofacModule.cs
```diff
@@ -69,6 +71,15 @@ private void RegisterConfigurations(ContainerBuilder builder)
 
         private void RegisterServices(ContainerBuilder builder)
         {
+            var connectionString = _configuration.GetConnectionString("Eurofurence");
+            var dbContextOptionsBuilder = new DbContextOptionsBuilder<AppDbContext>();
+            dbContextOptionsBuilder.UseMySql(connectionString, ServerVersion.AutoDetect(connectionString));
+
+            builder
```
Rain336, line 78:
```csharp
builder.AddDBContext<AppDbContext>(options =>
{
    var connectionString = _configuration.GetConnectionString("Eurofurence");
    options.UseMySql(connectionString, ServerVersion.AutoDetect(connectionString));
});
```

Rain336 (commented):
Looks good! Nothing big to complain, I guess! There are a few more topics, like request cancellation using `CancellationToken`, but they are not really specific to the db change, since nothing in the backend currently really does that, so I don't think they have to be done right now.

src/Eurofurence.App.Server.Services/DependencyResolution/AutofacModule.cs
Metawolve: Thanks for the suggestion, sadly in this case we are using autofac instead of the Microsoft pipeline so there is no AddDbContext. And the autofac module is used in two places (startup and the CLI toolbox) so I didn't want to refactor that yet. But I opened a separate issue to check if we can throw out autofac in favor of the Microsoft pipeline, so I would address this there

src/Eurofurence.App.Infrastructure.EntityFramework/AppDbContext.cs
Metawolve: Thanks, I implemented it in the latest commit!

src/Eurofurence.App.Server.Services/DependencyResolution/AutofacModule.cs
Rain336: Ah okay! Yeah, would be good since `AddDBContext` adds more than just the DB context, but that is for another time then! Looks good!

Rain336 (approved):

PR #43, opened 2024-05-06
Eurofurence Identity Integration

### LINKED ISSUE #36: Migrate to new Identity Provider
The backend should be migrated to the new identity provider:
https://identity.eurofurence.org/

The front-end apps will authenticate against the provider and send the access token to the backend which needs to validate the token, get the groups from the token, transform them into claims and authorize the users.

Currently we would only need two groups, admin and user.

For the new backoffice (the UI for the backend API), the authentication will probably be handled by the backend, using a second authentication provider that uses the user session instead of an access token to get the identity claims.

Rain336:
I ended up using the [`IdentityModel.AspNetCore.OAuth2Introspection`](https://github.com/IdentityModel/IdentityModel.AspNetCore.OAuth2Introspection) to implement the OAuth2 introspection based authentication. Not my favorite library, but it's the best that works for our use case. It even caches the introspection results, as long as the token is still valid.
Client id and endpoints can be configured in the `Identity` section of the `appsettings.json`
It cannot query the user info endpoint though, so I added that feature manually and merge the claims from the user info endpoint into the `ClaimsIdentity`.
~~Currently I configured the authorization policy so that every endpoint just requires a valid token, if not configured otherwise.~~ Endpoints require a `Authorize` attribute like always.
A few endpoints require cretin roles and they don't really map to anything right now, so calling these endpoints is currently impossible.
Closes #36


Metawolve:
Looks good so far! I think it's fine to use an existing library if we can't get it working well easily with the default Microsoft libraries.

Just one thing - I think we should rework the currently existing AuthenticationHandler or remove it if it is not needed anymore. There is also a "WhoAmI" endpoint, which I think won't work with the new authentication yet (correct me if I'm wrong).
Could you maybe check if this can be reworked or removed?
Other than that, it looks good to me!

Rain336:
I removed the `AuthenticationHandler` and anything related to the old token generation system, this also includes the telegram bot and toolbox commands for creating a RegSys Pin and displaying Pin usage information.

src/Eurofurence.App.Server.Services/Abstractions/Security/PrincipalExtensions.cs
Metawolve, line 1: Maybe we could set the principal name using the property "NameClaimType" along with the RoleClaimType in the ConfigureOAuth2IntrospectionOptions class? I know that the TokenValidationOptions of the JwtBearerOptions from Microsoft support this at least. This way we can set the Name property of the ClaimsPrincipal, instead of using an extension method. If it is not supported by the authentication library, then maybe in a claims transformation?

Rain336: The `NameClaimType` property exists, but it's default value is already `name`, so we don't need to change it.
I changed the code to use the `Name` property of the primary claims identity now.

Metawolve (approved):

PR #54, opened 2024-06-04
30 migrate from local disk storage to minio driven s3

Metawolve:
Before, images where saved as byte arrays to the database.
I have implemented minIO using the ofifcial nuget package to save the images to a minIO bucket. This is now the case for all images, including artist allay registration images and fursuit badge images.

The minIO connection and bucket can be configured in the appsettings.json, I have added an empty example into the appsettings.sample.json.

I have tested all affected API endpoints (images, artistAllay, fursuits) using Swagger and play.min.io, I also needed to adapt the CLIToolbox a bit and tested that too.


src/Eurofurence.App.Server.Services/Abstractions/MinIO/MinIOConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class MinIoConfiguration
+    {
+        public string Endpoint { get; set; }
+        public string AccessKey { get; set; }
+        public string SecretKey { get; set; }
+        public bool Secure { get; set; }
+        public string Bucket { get; set; }
+
+        public static MinIoConfiguration FromConfiguration(IConfiguration configuration)
```
Rain336, line 13: Instead of using this method, you can just let the configuration library bind the class automatically.
Just call `services.Configure<MinIoConfiguration>(configuration.GetSection("MinIo"))`, then you can inject the configuration as either `IOptionsSnapshot<MinIoConfiguration>` for transient services or `IOptionsMonitor<MinIoConfiguration>` for singleton services

Rain336 (changes requested):

src/Eurofurence.App.Server.Services/Abstractions/MinIO/MinIOConfiguration.cs
Metawolve: I wanted to do it that way, but I noticed that all other configurations use a FromConfiguration method like this, so I did it the same way for now. I think that the reason is, that we are using Autofac so IOptions with services.Configure doesn't work for the CLI toolbox for example.
So I did it the same way it was done in the other configurations for now, but I plan to replace this by the modern way of using IOptions for all configurations at some point, when we have removed Autofac

Rain336: Sounds good!

Rain336 (approved):

PR #88, opened 2024-07-12
37 new backoffice web UI as blazorrazor pages

Metawolve:
I added a blazor webassembly project with the new backoffice.
It includes OIDC logon with the Eurofurence identity provider, knowledge base management and image management.
The UI has been made using the free UI library Mudblazor (MIT license).

I also fixed a few minor issues with the API which needed to be fixed for some of the backoffice functionality.

The backoffice will be expanded further in separate issues.


Fenrikur (approved):
Missing validation on front- and backend (missing fields may result in 500), but lgtm otherwise.

PR #91, opened 2024-07-14
84 add api endpoint to create maps

Metawolve:
This pull request adds API endpoints for adding and deleting maps.
It also fixes some bugs when managing map entities.


Fenrikur (approved):
LGTM

PR #110, opened 2024-07-17
109 fix bugs from mudblazor 7 update

Metawolve:
Fixed some bugs caused by breaking changes in the new mudblazor version.
Fixed a minor bug when deleting images related to a map.


Fenrikur (approved):
LGTM

PR #128, opened 2024-07-27
fix(preview): missing icons and related entities

Fenrikur:
After the change to FA6 some icons were broken and as a leftover from the migration to EF, related entities need to be loaded explicitly.


Metawolve:
The changes from this pull requests have been copied to #130

PR #130, opened 2024-07-29
feat: add response and request models to hide relations from api

Metawolve:
I added request and response models for endpoints where it was needed to maintain a simpler API schema as it was before.
This includes knowledge base, images, maps and more.

Also the commits from #128 were included here.


Fenrikur (approved):
lgtm

PR #136, opened 2024-08-03
97 make fluentscheduler jobs out of imports

Metawolve:
I moved the imports of dealers, events, lost&found items and announcements into FluentScheduler jobs and made these and all other jobs configurable (see the appsettings.sample.json):

```
"jobs": {
  "flushPrivateMessageNotifications": {
    "enabled": true,
    "secondsInterval": 1
  },
  "updateAnnouncements": {
    "enabled": true,
    "secondsInterval": 60
  },
  "updateDealers": {
    "enabled": true,
    "secondsInterval": 60
  },
  "updateEvents": {
    "enabled": true,
    "secondsInterval": 60
  },
  "updateFursuitCollectionGameParticipation": {
    "enabled": true,
    "secondsInterval": 60
  },
  "updateLostAndFound": {
    "enabled": true,
    "secondsInterval": 60
  }
},
"lassie": {
  "baseApiUrl": "",
  "apiKey": ""
},
"dealers": {
  "url": "",
  "user": "",
  "password": ""
},
"announcements": {
  "url": ""
},
"events": {
  "url": "" // URL to statically hosted events csv export file
},
```

Note: I also squashed the entity framework migrations into one single migration because there were merge issues and migrations were getting a bit much.


src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -28,5 +57,232 @@ public IQueryable<EventRecord> FindConflicts(DateTime conflictStartTime, DateTim
...
+
+        public async Task RunImportAsync()
+        {
+            _logger.LogInformation(LogEvents.Import, "Starting events import.");
+
+            var httpClient = new HttpClient();
+            var fileStream = await httpClient.GetStreamAsync(_configuration.Url);
+            TextReader reader = new StreamReader(fileStream);
+
+            using var csv = new CsvReader(reader, new CsvConfiguration(CultureInfo.CurrentCulture) { Delimiter = "," });
```
Rain336, line 69: I don't know if `CultureInfo.CurrentCulture` is a good idea here, since it means, use the computer's language to parse stuff and I think that it affects parsing the `TimeSpan`s in `EventImportRow`. Using `CultureInfo.InvariantCulture` should make it use US-like style which is good enough for just the `TimeSpan`s.

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -28,5 +57,232 @@ public IQueryable<EventRecord> FindConflicts(DateTime conflictStartTime, DateTim
...
+                    : record.ConferenceDayName.Trim();
+
+            var conferenceTracks = csvRecords.Select(a => a.ConferenceTrack)
+                .Distinct().OrderBy(a => a).ToList();
+
+            var conferenceRooms = csvRecords.Select(a => a.ConferenceRoom)
+                .Distinct().OrderBy(a => a).ToList();
+
+            var conferenceDays = csvRecords.Select(a =>
+                    new Tuple<DateTime, string>(DateTime.SpecifyKind(DateTime.Parse(a.ConferenceDay), DateTimeKind.Utc),
```
Rain336, line 89: Similar case here, `DateTime.Parse` uses the current culture, so if you are running on a German computer it parses `dd.MM.yyyy HH:mm:ss` and on a US computer `MM/dd/yyyy hh:mm:ss`. Since the format is not going to change, using `DateTime.ParseExact` with a format string should work. (if it's an ISO 8601 date, you can just use `o` as the format string)

src/Eurofurence.App.Tools.CliToolBox/Importers/EventSchedule/CsvFileImporter.cs
```diff
@@ -166,13 +167,11 @@ IEventConferenceTrackService eventConferenceTrackService
         public int ImportCsvFile(string inputPath)
         {
             var stream = new FileStream(inputPath, FileMode.Open);
-            TextReader r = new StreamReader(stream);
+            TextReader reader = new StreamReader(stream);
 
-            var csv = new CsvReader(r, CultureInfo.CurrentCulture);
+            using var csv = new CsvReader(reader, new CsvConfiguration(CultureInfo.CurrentCulture) { Delimiter = "," });
```
Rain336, line 172: Same code as above

Rain336 (changes requested):

Rain336 (approved):

PR #140, opened 2024-08-04
137 fix soft delete

Metawolve:
This PR adds a global query filter to filter soft-deleted (isDeleted: 1) entities by default.


Fenrikur (approved):
lgtm

PR #141, opened 2024-08-04
fix: use jsonignore attribute instead of response classes

Metawolve:
This PR fixes the API schema using the JsonIgnore attribute.
Response types were removed where properties could be hidden with attributes.
Also, the delta response of the sync endpoint was fixed aswell, it's schema is now almost identical to last year (almost because there were some actual chagnes like the image url or knowledge group icon name).


Fenrikur (approved):
lgtm

Issue #143, opened 2024-08-06
Rework Access Roles


Metawolve:
From @Fenrikur :

Just some thoughts on the RBAC for the mobile app backend, going by actual roles we currently have at EF and might be worth seeing reflected in our RBAC system.

In addition to the regular, unprivileged endpoints:

- **Attendees** (authenticated and valid reg ID) should be able to:
  - register for and check Artist Alley registration,
  - retrieve private messages (incl status & count) and mark them as read
  - post event feedback (currently unprivileged)
  - retrieve list of lost & found items
- **Art Show** should be able to:
  - access all endpoints of the ArtShowController
- **Private Message Sender** (e.g. Photoshoot, Security, Art Show,…) should be able to:
  - send private messages (duh x'3)
  - list messages sent by them (think that's what :sent-by-me is for?)
  - check status of sent messages (is probably part of the result above)
- **Knowledge Base Editors** should be able to:
  - edit knowledge base groups and entries (CRUD)
  - but should not be able to delete/update arbitrary, non-KB images (so editing an image on a KB entry could just mean replacing it with a new one by POSTing the image and PUTting the KB entry)
- **Map Editors** should be able to:
  - edit map entries (CRUD)
- The following should be restricted to **Admins** (currently Developers?):
  - all write operations on announcements
  - (optional) CRUD on Artist Alley registrations (at least list and delete would be nice)
  - all write operations on dealers (maybe we will have a separate Dealer Admin role in the future, such that the Dealers' Den Regsys can directly publish updates)
  - delete & update images (otherwise editors could e.g. overwrite dealer images etc.)
  - create and change maps
  - privileged operations on push notifications

So we'd end up with the following roles:
- Attendee,
- Art Show Admin,
- Private Message Sender,
- Knowledge Base Editor,
- Map Editor and
- Admin

(yes, I left out the Fursuit Badge topic entirely since that will be handled by a system of its own and can likely be dropped in the future)

PR #145, opened 2024-08-07
feat(rbac): rework access roles

Metawolve:
The roles and their access were reworked according to the guideline (see #143).

Also, new API endpoints restricted to Admins were added, for querying a single artist alley table registration and deleting a table registration.


Rain336 (approved):

Fenrikur (approved):
lgtm

Issue #160, opened 2024-08-17
Manage Artist Alley Registrations in Backoffice


Fenrikur:
In order to be able to get an overview of all Artist Alley table registrations of the current year, there should be an overview in the Backoffice accessible to users with the `Admin` or the `ArtistAlleyAdmin` role, where they can view and also manage (accept, reject, delete) table registrations.

* List view can be built similar to the one used for images
* List view supports filtering by status and searching by display and owner name
* By default, list view is filtered to pending
* List view is paginated with a sensible number of entries per page (e.g. 20)
* List view shows `OwnerUsername`, `DisplayName`, `Location`, `ShortDescription` and an image thumbnail per row along with buttons to accept, reject and delete the row
* Detail view accessible from each row shows full details along with the same action buttons
* Actions prompt user for confirmation
* Status change is handled via a PUT endpoint `/Api/ArtistAlley/{id}/:status` which only accepts a status as body and is restricted to `Admin` and `ArtistAlleyAdmin` roles with only these valid tranisitions:
  * `Pending` -> `Rejected`
  * `Pending` -> `Accepted`
* `ArtistAlleyAdmin` role is configurable via `appsettings.json`

PR #178, opened 2024-08-25
feat(backoffice): manage artist alley registrations (#160)

Metawolve:
This PR adds artist alley management to the backoffice, introducing a new API endpoint for setting the artist alley table registration status, aswell as two new access roles, solving #160 .

It also adds a confirm dialog when trying to delete entities.

Thanks to @maakinoh for building the artist alley management :)


Fenrikur (approved):
Some minor things still need fixing, but feature looks good in general! Thanks @maakinoh  and @Metawolve! ♥ 
Fixes will happen within our actual repo. (=

PR #186, opened 2024-08-28
fix(push): Allow registering devices without login and re-registration

Rain336:
Allows adding device without an IdentityId and re-registering devices under a new IdentityId

### DIFF
--- a/src/Eurofurence.App.Server.Services/PushNotifications/FirebaseChannelManager.cs
+++ b/src/Eurofurence.App.Server.Services/PushNotifications/FirebaseChannelManager.cs
@@ -21,7 +21,7 @@ namespace Eurofurence.App.Server.Services.PushNotifications
     {
         private readonly IDeviceIdentityService _deviceService;
         private readonly IRegistrationIdentityService _registrationService;
-        private readonly AppDbContext _db;
+        private readonly AppDbContext _appDbContext;
         private readonly FirebaseConfiguration _configuration;
         private readonly ConventionSettings _conventionSettings;
         private readonly FirebaseApp _firebaseApp;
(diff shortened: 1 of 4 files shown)


src/Eurofurence.App.Server.Services/PushNotifications/FirebaseChannelManager.cs
```diff
@@ -206,17 +206,27 @@ public async Task RegisterDeviceAsync(
...
+                .AsTracking()
+                .ToListAsync(cancellationToken);
+
+            if (existing.Count > 0)
             {
-                return;
+                foreach (var identity in existing)
+                {
+                    identity.IdentityId = identityId;
+                }
```
Fenrikur, line 219: Since we're modifying tracked entities, we should probably `await _appDbContext.SaveChangesAsync(cancellationToken);` to write those changes to the db?

src/Eurofurence.App.Server.Services/PushNotifications/FirebaseChannelManager.cs
```diff
@@ -206,17 +206,27 @@ public async Task RegisterDeviceAsync(
...
-            }, cancellationToken);
+                await _deviceService.InsertOneAsync(new DeviceIdentityRecord
+                {
+                    IdentityId = identityId,
+                    DeviceToken = deviceToken,
+                    DeviceType = type
+                }, cancellationToken);
+            }
 
             var set = new HashSet<string>(regSysIds);
```
Fenrikur, line 231: We could just return right here if `regSysIds` is empty.

src/Eurofurence.App.Server.Services/PushNotifications/FirebaseChannelManager.cs
```diff
@@ -20,6 +21,7 @@ public class FirebaseChannelManager : IFirebaseChannelManager
     {
         private readonly IDeviceIdentityService _deviceService;
         private readonly IRegistrationIdentityService _registrationService;
+        private readonly AppDbContext _db;
```
Fenrikur, line 24: Please use `_appDbContext` instead of `_db` as that's how it's referred to in all other services. (=

Fenrikur (approved):
lgtm!

Metawolve (approved):

Fenrikur (approved):

PR #188, opened 2024-08-28
feat!: Removed previously deprecated image content endpoints

Metawolve:
Removed the image content endpoints.

Images have to be downloaded using the downloadUrl of the image response.


Fenrikur (approved):
lgtm

PR #189, opened 2024-08-29
Fix artist alley status filter and snackbar notification

maakinoh:
This PR is another small fix for #160.

Filtering by status is now possible using a custom FilterTemplate.

Moreover, the snackbar notifications when changing the applications' status are now having the same severity


Metawolve (approved):
lgtm

maakinoh:
This PR should not be merged but rather #190 since it used not my old fork.

The changes are the same

PR #190, opened 2024-08-29
 feat(backoffice): artist alley status filter and snackbar notification

maakinoh:
This PR is another small fix for #160.

Filtering by status is now possible using a custom FilterTemplate.

Moreover, the snackbar notifications shown when changing the applications' status are now having the same severity


Metawolve (approved):
lgtm

Fenrikur (approved):

PR #191, opened 2024-08-29
fix(pm): fix crash when reading private messages

Rain336:
Fixes a crash when reading privates messages, by changing how `PrivateMessageService` reads pending private messages and disabling `DBContext` pooling

### DIFF
--- a/src/Eurofurence.App.Server.Web/Startup.cs
+++ b/src/Eurofurence.App.Server.Web/Startup.cs
@@ -173,7 +173,7 @@ namespace Eurofurence.App.Server.Web
 
             services.Configure<LoggerFilterOptions>(options => { options.MinLevel = LogLevel.Trace; });
 
-            services.AddDbContext<AppDbContext>(options =>
+            services.AddDbContextPool<AppDbContext>(options =>
             {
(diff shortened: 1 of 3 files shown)


src/Eurofurence.App.Server.Web/Startup.cs
```diff
@@ -173,7 +173,7 @@ public void ConfigureServices(IServiceCollection services)
 
             services.Configure<LoggerFilterOptions>(options => { options.MinLevel = LogLevel.Trace; });
 
-            services.AddDbContextPool<AppDbContext>(options =>
+            services.AddDbContext<AppDbContext>(options =>
```
Fenrikur, line 176: Why would you replace the DbContextPool with a single DbContext? We can expect a _very_ high number of concurrent requests once the apps go live, so a proper pool sounds like the better solution.

Fenrikur (changes requested):

src/Eurofurence.App.Server.Web/Startup.cs
Rain336: After testing with the pool enabled again, I didn't notice any problems with the private message endpoint, so re-enabling it

Fenrikur (approved):

PR #193, opened 2024-08-31
fix(roles): remove spaces from regsys api response

Fenrikur:
The regsys status `checked in` did not get correctly mapped to the enum due to the space, which would likely apply to any other potential similar status in the future (e.g. `partially paid`). Replacing all spaces in returned status to fix this.


maakinoh (approved):
lgtm

PR #194, opened 2024-08-31
fix(dealers): exclude unwanted fields from API

Fenrikur:
* `AttendeeNickname` should only be exposed via `DisplayNameOrAttendeeNickname` if `DisplayName` is not set.
* `RegistrationNumber` should _never_ be exposed.


maakinoh (approved):
`JsonIgnore` will result in an database when creating new dealer entries error since both fields are still set as Required.

Therefore, the POST API endpoint `/Api/Dealers` cant be used anymore for creating new dealer entries.

Since dealers are created in a separate job this should still be fine to merge but the POST endpoint `/Api/Dealers` could be documented accordingly

maakinoh (approved):

PR #195, opened 2024-08-31
feat(artistalley): move input validation to service

Fenrikur:
* request fields now get properly validated in one place (the service)
* image only gets created/stored if all validations succeed first
* only display name, description, location and image are mandatory

### DIFF
--- a/src/Eurofurence.App.Server.Services/ArtistsAlley/TableRegistrationService.cs
+++ b/src/Eurofurence.App.Server.Services/ArtistsAlley/TableRegistrationService.cs
@@ -126,7 +126,7 @@ namespace Eurofurence.App.Server.Services.ArtistsAlley
                 DisplayName = _htmlSanitizer.Sanitize(request.DisplayName),
                 WebsiteUrl = request.WebsiteUrl,
                 ShortDescription = _htmlSanitizer.Sanitize(request.ShortDescription),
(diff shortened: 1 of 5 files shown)


src/Eurofurence.App.Server.Services/ArtistsAlley/TableRegistrationService.cs
```diff
@@ -97,7 +126,7 @@ public async Task RegisterTableAsync(ClaimsPrincipal user, TableRegistrationRequ
                 DisplayName = _htmlSanitizer.Sanitize(request.DisplayName),
                 WebsiteUrl = request.WebsiteUrl,
                 ShortDescription = _htmlSanitizer.Sanitize(request.ShortDescription),
-                TelegramHandle = request.TelegramHandle,
+                TelegramHandle = request.TelegramHandle.TrimStart('@'),
```
maakinoh, line 129: Sending a request with no telegram handle will result in a 500 err.
A null check is probably needed here
```suggestion
                TelegramHandle = request.TelegramHandle?.TrimStart('@'),
```

maakinoh (changes requested):
Just some minor issues are need to be addressed here.

Otherwise: lgtm

maakinoh (approved):
lgtm now

PR #213, opened 2024-09-03
feat: Made links editable

Metawolve:
Made links in knowledge entries editable.
Also fixed a bug where canceling the link fragment dialog would still change the link.


maakinoh (approved):
lgtm

PR #219, opened 2024-09-05
feat(push): use APNs instead of FCM for Apple devices

Fenrikur:
* introduce dotAPNS for APNs implementation
* rework push to use APNs directly instead of FCM since `expo-notifications` handles it that way
* prune invalid APNs device tokens after push attempts
* remove unused `FirebaseConfiguration.FirebaseTopics`
* move `firebase` and `apns` config into `push`
* sort packages in `csproj`` files


lukashaertel (approved):
Method names in csharp are uppercase, other than that lg

PR #221, opened 2024-09-06
fix(announcement): POST required Id and LastChangeDateTimeUtc

Fenrikur:
* POST endpoint required both `Id` and `LastChangeDateTimeUtc` to be set in the request instead of being set server-side.
* Improve error handling to avoid 500 errors with stacktraces.


src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -59,23 +69,45 @@ public async Task<ActionResult> DeleteAnnouncementAsync([FromRoute] Guid id)
         }
```
maakinoh, line 69: Maybe a 204 status should be returned here instead of a 200, when no body is returned

maakinoh (changes requested):
I found just a minor issue.

The rest looks good to me

maakinoh (approved):
lgtm

PR #228, opened 2024-09-11
fix(push): concurrency issues within PushNotificationChannelManager

Fenrikur:


src/Eurofurence.App.Server.Services/Abstractions/Communication/IPrivateMessageQueueService.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Eurofurence.App.Domain.Model.Communication;
```
maakinoh, line 6: These imports are not needed

src/Eurofurence.App.Server.Services/Communication/PrivateMessageQueueService.cs
```diff
@@ -0,0 +1,30 @@
...
+            lock (_notificationQueue)
+            {
+                if (_notificationQueue.TryDequeue(out var item))
+                    return item;
+                return default;
+            }
+        }
+
+        public int GetQueueSize() {
+            return _notificationQueue.Count;
```
maakinoh, line 27: This method should be in a lock too

maakinoh (changes requested):
Just two minor issues need to be addressed here

maakinoh (approved):

PR #230, opened 2024-09-13
feat(auth): support API keys

Fenrikur:


maakinoh (approved):
lgtm

src/Eurofurence.App.Server.Web/Identity/ApiKeyAuthenticationOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Eurofurence.App.Server.Web.Identity;
+
+public class ApiKeyAuthenticationOptions : AuthenticationSchemeOptions
+{
+    public IList<ApiKeyOptions> ApiKeys { get; set; }
+    public class ApiKeyOptions {
+        public string Key { get; set;}
+        public string PrincipalName { get; set;}
+        public DateTime ValidUntil { get; set;}
```
lukashaertel, line 13: Probably not an issue but that's a local time

lukashaertel (approved):

src/Eurofurence.App.Server.Web/Identity/ApiKeyAuthenticationOptions.cs
Fenrikur: Thanks for the hint; right now having this be precise enough on a day-level is sufficient though. (=

PR #249, opened 2024-11-22
40 remove autofac

Metawolve:
Removed autofac, refactored the startup to use the default DI pipeline and the minimal hosting model introduced in .NET 6 with extension methods.


src/Eurofurence.App.Server.Services/Abstractions/JobOption.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Eurofurence.App.Server.Services.Abstractions
+{
+    public class JobOption
+    {
+        public bool Enabled { get; set; }
+
```
maakinoh, line 6: Extra line here

maakinoh (approved):
Looks good to me.

Just some minor issues

maakinoh:
Not really an issue due to #250. But the Program.cs file of the CLI module is still importing Autofac

src/Eurofurence.App.Server.Services/Abstractions/ArtistsAlley/ArtistAlleyOptions.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Eurofurence.App.Server.Services.Abstractions.ArtistsAlley
+{
+    public class ArtistAlleyOptions
+    {
+        public bool RegistrationEnabled { get; set; }
+        public ArtistAlleyTelegramOptions Telegram { get; set; }
```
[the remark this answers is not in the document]
Metawolve: Sounds good, changed them to init-only and checked that they still work^^

src/Eurofurence.App.Server.Services/Abstractions/JobOption.cs
Metawolve: Removed

Metawolve:
Autofac is still imported in the CLItools and they currently don't work without it so they don't compile after this PR. But due to #250 I think we don't need to put in the time to make that work.

PR #265, opened 2024-11-28
refactor: format code and add husky to project

maakinoh:
This PR will add the husky tool to this project, which is used for managing git hooks.

Currently, husky will automatically format any staged files before each commit with `dotnet format`.
This would allow us to have a more consistent code style and will also lower the risk of merge conflicts.


Metawolve (approved):
lgtm

PR #269, opened 2024-12-24
fix: docker build failed

maakinoh:
The docker build has been failing recently.
Probably due to several PRs (#242 #250,and #265) - some of which I created myself, to my shame.

+ The Clitoolbox was still referenced at some places
+ Husky.Net wasn't disabled in docker properly.
+ We've updated to dotnet 9 too early, so I rolled back to version 8, since we don't target version 9 yet.
Correct me if I'm wrong here @Metawolve, but AFAIK, you can only compile a dotnet project with the same version of dotnet installed, right?
+ Transitive dependency caused a compile error and I had to upgrade it (no breaking changes from what I'm seeing).


Metawolve (approved):
lgtm

Issue #276, opened 2025-01-12
Making Event favorites a server-side feature


Metawolve:
Favorites can be made server-side so we have more options for new features in the future, like filtering notifications for favorites only.

Idea/Nice-To-Have: Allow sharing favorite events with others, so attendees can see in the event overview if friends 
attend an event

PR #284, opened 2025-02-09
feat: mark events as favorites

maakinoh:
Closes: #276.

This PR introduces new endpoints and a database table for marking events as favorites.

The nice to have feature about 'sharing favorite events' is probably best handled in a separate PR.


Metawolve (changes requested):
This looks like a reasonable solution.
But maybe a suggestion - not sure which way is actually better - but another way would be to have a direct n:m relationship between users and events. Meaning that the EventRecord would have a new property that is a list of users (e.g. "FavoredBy") and the UserRecord will have a new property that is a list of events (e.g. "FavoriteEvents"). The names are just suggestions, there's probably a better way to name them.
But this would mean that you wouldn't need an EventFavoriteRecord class, as EF Core will handle this automatically and create a relation table in the database (probably called UserRecordsEventRecords or something, although you can customize the name in the modelbuilder).
You then can just add events to favorites by just adding them to the list of FavoriteEvents of the user, and remove them by just removing them from that list.
Then you also could easily get the list of users that marked an event as favorite, by just getting event.FavoredBy, e.g. to show who favored an event in the app, if we want to do that in the future.

maakinoh:
Good point, your approach would be more clean.

I wasn't quite sure if the `UserRecord` was still in active use, apart from the TG bot, and actively filled with data.
But I think it shouldn't be an issue to add the users' IDP Id to this table then, right? Dunno if this approach would mess somewhere.

src/Eurofurence.App.Server.Services/Events/EventFavoritesService.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public class EventFavoritesService(
+        AppDbContext appDbContext,
+        IStorageServiceFactory storageServiceFactory,
+        bool useSoftDelete = true)
+        : IEventFavoritesService
+    {
+        public async Task AddEventToFavoritesIfNotExist(ClaimsPrincipal user, EventRecord eventRecords)
+        {
+            if (!appDbContext.Users.Any(x =>
```
Metawolve, line 23: The outer if statement could be removed, instead we could use FirstOrDefault and throw an error if the user is not found

src/Eurofurence.App.Server.Services/Events/EventFavoritesService.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Returns a list with all favorite events of a given user by their regid : <paramref name="user"/>
+        /// </summary>
+        /// <param name="user">The user whose events should be fetched</param>
+        /// <returns>A list of all the events of the user</returns>
+        public List<EventRecord> GetFavoriteEventsFromUser(ClaimsPrincipal user)
+        {
```
Metawolve, line 45: Same thing, you can use .FirstOrDefault checking if the user exists, and if it does, just return foundRecord.FavoriteEvents

src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
```diff
@@ -66,6 +73,59 @@ public async Task<EventRecord> GetEventAsync([FromRoute] Guid id)
...
+        /// <summary>
+        /// Adds an event to favorites
+        /// </summary>
+        /// <param name="id">The id of the event</param>
+        /// <returns>Just a status code</returns>
+        [Authorize]
+        [HttpPost("{id}/:favorite")]
+        public async Task<ActionResult> MarkEventAsFavorite([FromRoute] Guid id)
+        {
+            var events = _eventService.FindAll(x => x.Id == id).AsNoTracking();
```
Metawolve, line 97: You can just use the method _eventService.FindOneAsync here, that finds the event by id, and already does the .AsNoTracking for you

src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
```diff
@@ -66,6 +73,59 @@ public async Task<EventRecord> GetEventAsync([FromRoute] Guid id)
...
+        public async Task<ActionResult> UnmarkEventAsFavorite([FromRoute] Guid id)
+        {
+            var events = _eventService.FindAll(x => x.Id == id);
+
+            if (!events.Any())
+            {
+                return NotFound();
+            }
+
+            await _eventFavoriteService.RemoveEventFromFavoritesIfExist(User, events.First());
```
Metawolve, line 124: Maybe we could unify the method names in the controllers and services a bit, so it is more obvious that they belong together.
We could also think about putting the event favorite related stuff into the eventService instead of it having it's own service, as they are sharing the same controller aswell. Although that's matter of taste I think :)

Metawolve (commented):
Looks good apart from some minor comments

src/Eurofurence.App.Server.Services/Events/EventFavoritesService.cs
maakinoh: I think the `user` object is never null since we use this method in an Authorized endpoint.
But a sanity check can't hurt.

src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
maakinoh: That's a good idea.
I just left it that way because of the old database schema, but that way it makes more sense with the new db schema.

Metawolve (approved):
Lgtm

Issue #277, opened 2025-01-12
New API endpoint for updating calendar entries for favorite events


Metawolve:
Authenticated API endpoint to subscribe to events as ICS for favorite events, so they get updated in the calendar automatically.

Depends on #276

PR #293, opened 2025-03-11
feat(#277): add favorite events calendar endpoint

maakinoh:
Closes: #277 

This PR adds two new endpoints:

- Access favorite events in ICS format.
- Acquire an access token for the ICS endpoint.


src/Eurofurence.App.Infrastructure.EntityFramework/Migrations/20250225212508_AddUserCalendarToken.cs
```diff
@@ -0,0 +1,72 @@
...
+                name: "IX_Events_TelegramUserRecordId",
+                table: "Events");
+
+            migrationBuilder.DropColumn(
+                name: "RegsysID",
+                table: "TelegramUsers");
+
+            migrationBuilder.DropColumn(
+                name: "TelegramUserRecordId",
+                table: "Events");
```
Metawolve, line 28: Are the dropped columns, indexes and the dropped foreign key in the migration correct? Seems like changes from the last PR regarding the user rework are reverted with this and I don't see changes on the TelegramUserRecord.cs in this PR.

Metawolve (changes requested):

Metawolve (approved):
Looks good to me

Issue #302, opened 2025-03-23
Add artist alley feed endpoint


maakinoh:
There should be an endpoint for listing all approved artists alley registrations as a normal attendee.

For moderators and admins, there should still be an endpoint for listing all registrations (including unapproved).

PR #309, opened 2025-04-13
feat(#302): changed role restriction of artist alley feed endpoint

maakinoh:
Allow checked-in attendees to request all accepted artists.

Closes: #302

### DIFF
--- a/src/Eurofurence.App.Server.Web/Controllers/ArtistsAlleyController.cs
+++ b/src/Eurofurence.App.Server.Web/Controllers/ArtistsAlleyController.cs
@@ -25,7 +25,13 @@ namespace Eurofurence.App.Server.Web.Controllers
         private readonly IArtistAlleyUserPenaltyService _artistAlleyUserPenaltyService;
         private readonly ArtistAlleyOptions _artistAlleyOptions;
 
-        public ArtistsAlleyController(ITableRegistrationService tableRegistrationService, IOptions<ArtistAlleyOptions> artistAlleyOptions, IImageService imageService, IArtistAlleyUserPenaltyService artistAlleyUserPenaltyService)
+        private const string AdminRoleName = "Admin";
+        private const string ArtistAlleyModeratorName = "ArtistAlleyModerator";
+        private const string ArtistAlleyAdminName = "ArtistAlleyAdmin";
+
+        public ArtistsAlleyController(ITableRegistrationService tableRegistrationService,
+            IOptions<ArtistAlleyOptions> artistAlleyOptions, IImageService imageService,
+            IArtistAlleyUserPenaltyService artistAlleyUserPenaltyService)
         {
             _tableRegistrationService = tableRegistrationService;
             _imageService = imageService;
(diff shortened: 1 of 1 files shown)


src/Eurofurence.App.Server.Web/Controllers/ArtistsAlleyController.cs
```diff
@@ -25,7 +25,13 @@ public class ArtistsAlleyController : BaseController
         private readonly IArtistAlleyUserPenaltyService _artistAlleyUserPenaltyService;
         private readonly ArtistAlleyOptions _artistAlleyOptions;
 
-        public ArtistsAlleyController(ITableRegistrationService tableRegistrationService, IOptions<ArtistAlleyOptions> artistAlleyOptions, IImageService imageService, IArtistAlleyUserPenaltyService artistAlleyUserPenaltyService)
+        private const string AdminRoleName = "Admin";
```
Metawolve, line 28: I think these constants are now unused and can be removed, right?

src/Eurofurence.App.Server.Web/Controllers/ArtistsAlleyController.cs
```diff
@@ -60,17 +66,33 @@ public async Task<ActionResult> PutTableRegistrationStatusAsync([EnsureNotNull][
...
+        [Authorize(Roles = "Admin, ArtistAlleyModerator, ArtistAlleyAdmin")]
+        [HttpGet("all")]
+        public IEnumerable<TableRegistrationRecord> GetAllTableRegistrationRecords()
+        {
+            return _tableRegistrationService.GetRegistrations(null);
+        }
+
         /// <summary>
         ///     Retrieves a list of all table registrations.
+        ///     Pending or rejected registrations are only visible to admins and moderators.
```
Metawolve, line 85: As this endpoint does never returns rejected)/ not accepted registrations now, the comments  should be changed accordingly

src/Eurofurence.App.Server.Web/Controllers/ArtistsAlleyController.cs
```diff
@@ -60,17 +66,33 @@ public async Task<ActionResult> PutTableRegistrationStatusAsync([EnsureNotNull][
...
+        ///     Checked-in attendees may only see accepted registrations.
         /// </summary>
         /// <returns>All table registrations.</returns>
         [ProducesResponseType(typeof(string), 404)]
         [ProducesResponseType(typeof(IEnumerable<TableRegistrationRecord>), 200)]
-        [Authorize(Roles = "Admin, ArtistAlleyModerator, ArtistAlleyAdmin")]
+        [Authorize(Roles = "AttendeeCheckedIn")]
```
Metawolve, line 91: I think we can give Admin, ArtistAlleyModerator and ArtistAlleyAdmin access to this endpoint aswell. They will probably use the all-endpoint, but I don't see a reason why they wouldn't be allowed to access this endpoint aswell.

Metawolve (commented):
Had some minor comments, apart from that it looks good to me

src/Eurofurence.App.Server.Web/Controllers/ArtistsAlleyController.cs
maakinoh: I assumed that Admins would have the AttendeeCheckedIn role anyway, but it is probably better your way.

Metawolve (approved):
Looks good now

PR #321, opened 2025-06-10
318 efnav integration

Metawolve:
Added property "MapLink" to dealers responses and event conference room responses, which is computed at runtime for all affected endpionts including the sync endpoint.

The template for the map link is configured via the appsettings (added to appsettings.sample.json):

```
"Maps": {
  "UrlTemplate": "https://nav.eurofurence.org/l/uuid-{id}/"
}
```


maakinoh (approved):
LGTM

PR #322, opened 2025-06-15
Backend Sentry Integration

### LINKED ISSUE #311: Backoffice | Sentry integration
We would like to integrate Sentry for the Backoffice, as we are using it for the app aswell.

Rain336:
This PR adds Sentry to the Backend as well as the Backoffice.
The DSN to connect to can be configured in the respective `appsettings.json`
In a non-production environment (running the Backend locally, for example) Pii is sent to sentry too and the sample rate is set to 100% for testing purposes.
I also wrote a `ISentryUserFactory` that adds a User's IdentityId, Username, Email, IP (Pii) and Roles to Sentry traces.
For the recurring jobs, I added a `IJobListener` to Quarz to track sucessful and failing jobs in Sentry.
Closes #311


Metawolve (approved):
LGTM

PR #327, opened 2025-06-29
fix(logging): init early & consistently use Serilog

Fenrikur:
- Logging from Quartz jobs did not work
- Initial log lines would not adhere to formatting
- Duplicate console logger initialized


Metawolve (approved):
LGTM

PR #333, opened 2025-07-16
fix(knowledge): non-published entries in sync endpoint

Metawolve:
Fixed a bug where non-published knowledge base entries would show up in the sync endpoint response


Fenrikur (commented):
tested locally and lgtm!

Fenrikur (approved):

PR #334, opened 2025-07-17
fix(backoffice): workaround for NullabilityInfoContext_NotSupported

Fenrikur:
see https://github.com/dotnet/aspnetcore/issues/59806

### DIFF
--- a/src/Eurofurence.App.Server.Web/appsettings.sample.json
+++ b/src/Eurofurence.App.Server.Web/appsettings.sample.json
@@ -12,19 +12,21 @@
   },
   "Identity": {
     "ClientId": "",
+    "GroupsEndpoint": "https://identity.eurofurence.org/api/v1/groups",
     "IntrospectionEndpoint": "https://identity.eurofurence.org/api/v1/introspect",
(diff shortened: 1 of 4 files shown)


src/Eurofurence.App.Server.Web/appsettings.sample.json
```diff
@@ -12,19 +12,21 @@
...
-    "GroupsEndpoint": "https://identity.eurofurence.org/api/v1/groups",
     "RegSysUrl": ""
   },
   "Authorization": {
+    "Admin": [],
     "ArtistAlleyAdmin": [],
     "ArtistAlleyModerator": [],
     "ArtShow": [],
+    "Attendee": [],
+    "AttendeeCheckedIn": [],
```
Metawolve, line 26: This is not something we fill from the appsettings, but from the reg system if I'm not mistaken.

Fenrikur: Can’t check the code rn to provide a reference, but these settings should allow devs to test these roles without the regsys by assigning them via group. Pretty sure we added this a while ago and it’s actually how you can get these roles on the test system. (;

Metawolve (approved):
LGTM

PR #337, opened 2025-07-30
fix(model): Mapster config for KnowledgeEntries

Fenrikur:
* Fixes an issue where KnowledgeEntries were not correctly mapped to their respective response objects resulting in missing data on relations.
* Includes some DRY optimisations on the KnowledgeEntryService.


Metawolve (approved):
LGTM

PR #347, opened 2025-08-09
feat(events): internal events accessible for staff only

Fenrikur:
- allows defining event tracks as internal
- adds internal flag to all event related entities
- adds new Staff role
- filters internal event entities from all unauthenticated, non-staff endpoints

### DIFF
--- a/src/Eurofurence.App.Server.Services/Events/EventService.cs
+++ b/src/Eurofurence.App.Server.Services/Events/EventService.cs
@@ -35,7 +35,7 @@ namespace Eurofurence.App.Server.Services.Events
         private readonly IEventConferenceRoomService _eventConferenceRoomService;
         private readonly IEventConferenceTrackService _eventConferenceTrackService;
         private readonly EventOptions _eventOptions;
-        private readonly AppDbContext _dbContext;
+        private readonly AppDbContext _appDbContext;
 
         private static readonly SemaphoreSlim Semaphore = new(1, 1);
 
@@ -52,7 +52,7 @@ namespace Eurofurence.App.Server.Services.Events
         )
             : base(appDbContext, storageServiceFactory)
         {
-            _dbContext = appDbContext;
+            _appDbContext = appDbContext;
             _eventConferenceDayService = eventConferenceDayService;
             _eventConferenceRoomService = eventConferenceRoomService;
             _eventConferenceTrackService = eventConferenceTrackService;
@@ -66,7 +66,7 @@ namespace Eurofurence.App.Server.Services.Events
         {
             if (user == null) throw new ArgumentNullException(nameof(user));
 
-            UserRecord userRecord = _dbContext.Users
+            UserRecord userRecord = _appDbContext.Users
                 .Include(userRecord => userRecord.FavoriteEvents)
                 .FirstOrDefault(x => x.IdentityId == user.GetSubject());
 
@@ -75,7 +75,7 @@ namespace Eurofurence.App.Server.Services.Events
                 userRecord.FavoriteEvents.Add(eventRecord);
             }
 
-            await _dbContext.SaveChangesAsync();
+            await _appDbContext.SaveChangesAsync();
         }
 
         /// <summary>
@@ -85,7 +85,7 @@ namespace Eurofurence.App.Server.Services.Events
         /// <returns>A list of all the events of the user</returns>
         public List<EventRecord> GetFavoriteEventsFromUser(ClaimsPrincipal user)
         {
-            return _dbContext.Users
+            return _appDbContext.Users
                 .AsNoTracking()
                 .Include(x => x.FavoriteEvents)
                 .Where(x => x.IdentityId == user.GetSubject())
@@ -119,7 +119,7 @@ namespace Eurofurence.App.Server.Services.Events
 
         public async Task RemoveEventFromFavoritesIfExist(ClaimsPrincipal user, EventRecord eventRecord)
         {
-            var foundRecord = _dbContext.Users
+            var foundRecord = _appDbContext.Users
                 .Include(x => x.FavoriteEvents)
                 .First(x => x.IdentityId == user.GetSubject());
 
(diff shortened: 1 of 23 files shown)


src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
```diff
@@ -138,12 +150,15 @@ public async Task<ActionResult> FavoritesCalendar([FromQuery, BindRequired] stri
         /// <param name="id">The id of the event</param>
         /// <returns>Just a status code</returns>
         [Authorize]
+        [AllowAnonymous]
+        [Authorize]
```
maakinoh, line 174: `[Authorize]` is dulicated and also in conflict with `[AllowAnonymous]` which does the exact oposite

src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
```diff
@@ -158,6 +173,8 @@ public async Task<ActionResult> MarkEventAsFavorite([FromRoute] Guid id)
...
```
maakinoh, line 202: Same problem with `[Authorize]` here.

src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
```diff
@@ -38,12 +38,15 @@ public EventsController(IEventService eventService,
         ///     Retrieves a list of all events in the event schedule.
         /// </summary>
         /// <returns>All events in the event schedule.</returns>
+        [Authorize]
+        [AllowAnonymous]
```
maakinoh, line 47: `[AllowAnonymous]` voids `[Authorize]`. Is this intentional?

src/Eurofurence.App.Server.Web/Controllers/EventConferenceTracksController.cs
```diff
@@ -24,24 +26,30 @@ public EventConferenceTracksController(IEventConferenceTrackService eventConfere
         ///     Retrieves a list of all event conference tracks in the event schedule.
         /// </summary>
         /// <returns>All events in the event schedule.</returns>
+        [Authorize]
+        [AllowAnonymous]
```
maakinoh, line 35: `[AllowAnonymous]` voids `[Authorize]`. Is this intentional?

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -164,93 +167,103 @@ public async Task RunImportAsync()
 
                 if (csvRecords.Count == 0) return;
 
+                var internalTrackNamesLowerCase = _eventOptions.InternalTracksLowerCase;
```
maakinoh: Maybe a null check is missing here.
If the entry is not in the config file a null pointer could be thrown.
```suggestion
                var internalTrackNamesLowerCase = _eventOptions.InternalTracksLowerCase ?? Array.Empty<string>();
```

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -101,6 +101,9 @@ public Calendar GetFavoriteEventsFromUserAsIcal(UserRecord user)
 
             foreach (var item in favoriteEvents)
             {
+                // TODO: Check user authorisation for internal events
```
maakinoh: Is this TODO already addressed?

src/Eurofurence.App.Server.Web/Controllers/EventConferenceDaysController.cs
```diff
@@ -24,24 +26,30 @@ public EventConferenceDaysController(IEventConferenceDayService eventConferenceD
         ///     Retrieves a list of all event conference days in the event schedule.
         /// </summary>
         /// <returns>All events in the event schedule.</returns>
+        [Authorize]
+        [AllowAnonymous]
```
maakinoh, line 35: Conflicting authentication rules here too.

src/Eurofurence.App.Server.Web/Controllers/EventConferenceDaysController.cs
```diff
@@ -24,24 +26,30 @@ public EventConferenceDaysController(IEventConferenceDayService eventConferenceD
...
+            var isStaff = User?.IsInRole("Staff") ?? false;
+            return _eventConferenceDayService.FindAll(x => isStaff || !x.IsInternal).Select(x => x.Transform());
         }
 
         /// <summary>
         ///     Retrieve a single event conference day in the event schedule.
         /// </summary>
         /// <param name="id">id of the requested entity</param>
+        [Authorize]
+        [AllowAnonymous]
```
maakinoh, line 55: Conflicting authentication rules here too.

src/Eurofurence.App.Server.Web/Controllers/EventConferenceRoomsController.cs
```diff
@@ -25,12 +26,15 @@ public EventConferenceRoomsController(IEventConferenceRoomService eventConferenc
         ///     Retrieves a list of all event conference Rooms in the event schedule.
         /// </summary>
         /// <returns>All events in the event schedule.</returns>
+        [Authorize]
+        [AllowAnonymous]
```
maakinoh, line 35: Conflicting authentication rules here too.

src/Eurofurence.App.Server.Web/Controllers/EventConferenceRoomsController.cs
```diff
@@ -39,12 +43,15 @@ public async Task<IEnumerable<EventConferenceRoomResponse>> GetEventsAsync()
         ///     Retrieve a single event conference room in the event schedule.
         /// </summary>
         /// <param name="id">id of the requested entity</param>
+        [Authorize]
+        [AllowAnonymous]
```
maakinoh, line 57: Conflicting authentication rules here too.

maakinoh (changes requested):
Thank you for this PR.
I just found some minor issues which should be addressed. 

The rest looks good to me.

src/Eurofurence.App.Server.Services/Events/EventService.cs
Fenrikur: No, that was added as a future todo and isn't meant to be addressed in this PR, since it's more of an academic edge case. The only scenario would be somebody adding a favourite while Staff and then still having access to it even after being removed from staff (for that year). It is likely something we don't really need to address at all and which would be difficult to check anyways as we would need to get the user's current groups from the IDP without having a token from them we could introspect.

tl;dr
Simply ignore for now please. (:

Metawolve (approved):
Last changed LGTM

maakinoh (approved):
LGTM

PR #352, opened 2025-08-17
feat(qr): make target IDs case-insensitive

Fenrikur:


maakinoh (approved):
Didn't test it myself, but it looks good to me.

PR #360, opened 2026-04-08
build: Updated all projects and dependencies to .NET 10 where possible

Metawolve:
Updates all projects to target .NET 10.
Updates most dependencies.
Updates .NET version in dockerfile.

Pomelo does not support .NET 10 yet, so we keep using .NET 8 for those libraries. Also fixed some issues that occoured due to breaking changes in OpenAPI 3 and MudBlazor update.


src/Eurofurence.App.Server.Web/Startup/ServiceCollectionExtensions.cs
```diff
@@ -289,11 +290,30 @@ public static IServiceCollection AddMapper(this IServiceCollection services)
...
             return services;
         }
+
+        public static IServiceCollection AddFirebase(this IServiceCollection services, FirebaseOptions firebaseOptions)
+        {
+            if (string.IsNullOrEmpty(firebaseOptions.GoogleServiceCredentialKeyFile))
+                return services;
+
+            var credential = CredentialFactory.FromFile<ServiceAccountCredential>(firebaseOptions.GoogleServiceCredentialKeyFile)
+                                  .ToGoogleCredential();
```
maakinoh, line 306: Maybe we could do a quick existence check of the file here and log an error if the `firebase.json` doesn't exist.

That way we get a clearer error message

maakinoh (changes requested):
Thank you for this PR.

I just have a small nitpick - the rest looks fine.

maakinoh (approved):

PR #362, opened 2026-04-10
Feat: #291 generate dmc from reg id

### LINKED ISSUE #291: Generate DataMatrixCode from Reg ID
We want a data matrix code in the app that cointains the reg number of the attendee.

We can generate this code in the backend, so we can at a later point add it to the Google/Apple wallet (that can be a separate issue).

maakinoh:
Adds a new endpoint for generating a data matrix code with the reg id.

This code should be the same as on the conbadge.

Closes #291


Metawolve (dismissed):
LGTM

Metawolve (approved):
LGTM

PR #371, opened 2026-05-13
Refactor/rename user code endpoint

maakinoh:
Closes: #370 

Renames the user/datamatrix endpoint to user/pass


Fenrikur (approved):
Approved to ensure we have the correct endpoint name.

Requires follow up work:
* No role requirement (any signed in user is permitted to call this method)
* Should return 404 if user has no valid registration.

PR #380, opened 2026-05-25
refactor: remove role requirement for Users/pass endpoint

maakinoh:
Closes #373 

Removes the role requirement for the Users/pass endpoint.
If the user has no active registration, a 404 is returned; otherwise, the latest registration is returned.

Furthermore, the image type of the matrix code can now be changed by the `Accept` header.


Metawolve (dismissed):
LGTM

Metawolve (approved):
LGTM

PR #381, opened 2026-05-26
fix: add new SecurityRequirement to swagger configuration

maakinoh:
With some swagger update, the configuration must have changed, and the Bearer token is not sent anymore.

According to the swagger docs, we need to add a new SecurityRequirement.


Metawolve (approved):
LGTM

PR #396, opened 2026-05-28
refactor: changed datamatrix image type specification

maakinoh:
The Accept header might not be appropriate for specifying the image type.

Now we should use a separate URL parameter for that.

Also, the endpoint name `/pass` was changed to `/Pass` for consistency.

### DIFF
--- a/src/Eurofurence.App.Server.Web/Controllers/UsersController.cs
+++ b/src/Eurofurence.App.Server.Web/Controllers/UsersController.cs
@@ -96,7 +96,7 @@ namespace Eurofurence.App.Server.Web.Controllers
         [HttpGet("pass")]
         [ProducesResponseType(typeof(FileContentResult), 200)]
         [ProducesResponseType(typeof(string), 404)]
-        public ActionResult GetDataMatrixCode([FromQuery] string imageType)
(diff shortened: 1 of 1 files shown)


src/Eurofurence.App.Server.Web/Controllers/UsersController.cs
```diff
@@ -92,26 +91,27 @@ public async Task<ActionResult> PutArtistAlleyUserPenaltyAsync([EnsureNotNull][F
...
+        /// <param name="imageType">The image type the resulting data matrix code should have.</param>
         /// <returns>Data matrix code as svg.</returns>
-        [HttpGet("pass")]
+        [HttpGet("Pass")]
         [ProducesResponseType(typeof(FileContentResult), 200)]
         [ProducesResponseType(typeof(string), 404)]
-        public ActionResult GetDataMatrixCode()
+        public ActionResult GetDataMatrixCode([Required][FromQuery] string imageType)
         {
             if (User.Identity is ClaimsIdentity identity && _identityService.GetRegistrationsIds(identity).Any())
```
Metawolve: Prefer early returns here.. Maybe using Unauthorized()

Metawolve (approved):
LGTM

PR #407, opened 2026-06-12
build(deps): Updated dependencies where sufficient

Metawolve:
- Updated dependencies where sufficient, fixing vulnarability warning and build issues.
- Replaced depcreated and vulnerable IdentityModel.AspNetCore.OAuth2Introspection with Duende.AspNetCore.Authentication.OAuth2Introspection, which is the rebranding
  -> This now uses the Asp.NET default caching by default, therefore added .AddDistributedMemoryCache and .AddHybridCache  in startup pipeline

Not updated:
- ImageSharp needs a community license for version 4.0.0, which has to be obtained from their website via a form (https://licensing.sixlabors.com/)
- Pomelo and base EntityFramework libraries can't be updated, because Pomelon currently does not have .NET 10 support yet


maakinoh (approved):
LGTM

PR #417, opened 2026-06-21
fix: replaces DataMatrix.NetCore with ZXing.Net

maakinoh:
Replaces DataMatrix.NetCore with ZXing.Net as the datamatrix generator lib.


Metawolve (approved):
LGTM. ZXing.Net looks like a better supported alternative

PR #425, opened 2026-06-25
feat(events): use schedule version for sync

Fenrikur:
* reduce load due to unnecessary event imports by checking the schedule version ahead of performing a full fetch & import
* expose latest imported schedule version via API


Metawolve (approved):
Thought about if we should use System.Version instead of string, but this is probably more failsafe, as we do not have to assume a valid version format and can just check if it's different from last time. So approved

PR #435, opened 2026-07-01
feat: add event feedback get endpoint

### LINKED ISSUE #428: GET Endpoint for Event feedback
Currently, event feedback is collected via SQL queries.

It would be nice to have a GET endpoint for this.
For all event feedback, with an optional event ID parameter for a single event.

Permissions would be for role Programming only.

maakinoh:
Closes #428 

Adds a new endpoint for getting all event feedback records filterable by an optional EventSourceId.

### DIFF
--- a/src/Eurofurence.App.Server.Web/Controllers/EventFeedbackController.cs
+++ b/src/Eurofurence.App.Server.Web/Controllers/EventFeedbackController.cs
@@ -55,18 +55,18 @@ namespace Eurofurence.App.Server.Web.Controllers
         /// <summary>
         /// Returns all event feedback records filterable by event id.
         /// </summary>
-        /// <param name="filterEventId">Optional event source id filter.</param>
+        /// <param name="eventSourceId">Optional event source id filter.</param>
         /// <returns>A list of all event's (or filtered) feedback.</returns>
         [Authorize(Roles = "EventFeedbackManager")]
         [HttpGet]
         [ProducesResponseType(typeof(IEnumerable<EventFeedbackResponse>), 200)]
-        public ActionResult GetEventFeedback(int? filterEventId)
+        public ActionResult GetEventFeedback(int? eventSourceId)
         {
             IEnumerable<EventFeedbackRecord> records = _eventFeedbackService.FindAll();
             EventFeedbackRecord first = records.FirstOrDefault();
-            if (filterEventId != null)
+            if (eventSourceId != null)
             {
-                records = records.Where(x => x.Event.SourceId == filterEventId);
+                records = records.Where(x => x.Event.SourceId == eventSourceId);
             }
             return Ok(records.Select(x => x.Transform()));
         }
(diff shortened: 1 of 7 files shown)


src/Eurofurence.App.Server.Web/Controllers/EventFeedbackController.cs
```diff
@@ -47,5 +51,24 @@ public async Task<ActionResult> PostEventFeedbackAsync([FromBody] PostEventFeedb
...
+        /// </summary>
+        /// <param name="eventSourceId">Optional event source id filter.</param>
+        /// <returns>A list of all event's (or filtered) feedback.</returns>
+        [Authorize(Roles = "EventFeedbackManager")]
+        [HttpGet]
+        [ProducesResponseType(typeof(IEnumerable<EventFeedbackResponse>), 200)]
+        public ActionResult GetEventFeedback(int? eventSourceId)
+        {
+            IEnumerable<EventFeedbackRecord> records = _eventFeedbackService.FindAll();
+            EventFeedbackRecord first = records.FirstOrDefault();
```
Metawolve: Unused variable?

src/Eurofurence.App.Server.Web/Controllers/EventFeedbackController.cs
```diff
@@ -47,5 +51,24 @@ public async Task<ActionResult> PostEventFeedbackAsync([FromBody] PostEventFeedb
 
             return NoContent();
         }
+
+        /// <summary>
+        /// Returns all event feedback records filterable by event id.
+        /// </summary>
+        /// <param name="eventSourceId">Optional event source id filter.</param>
+        /// <returns>A list of all event's (or filtered) feedback.</returns>
+        [Authorize(Roles = "EventFeedbackManager")]
```
Metawolve: Should we add Admin here too? Admins currently have access on everything except this, as far as I can see it.
Would be [Authorize(Roles = "Admin, EventFeedbackManager")]

src/Eurofurence.App.Server.Web/Controllers/EventFeedbackController.cs
maakinoh: Forgot to remove it. Already fixed

src/Eurofurence.App.Server.Web/Controllers/EventFeedbackController.cs
maakinoh: Good idea - fixed.

I think we have a few places (including the post method in this controller) where the Admin role is not set in the authorized role list.
If admins should be able to access everything (which would make sense), we probably need to do some refactoring.
Should I create a new issue for this?

Metawolve: Yes, probably not a big deal for now but would make sense to me to do that

Metawolve (approved):
LGTM

PR #437, opened 2026-07-07
feat(events): favorite statistics based on reg status

### LINKED ISSUE #294: Backoffice | View event favorite count
There should be a view in the backoffice in which to see how often an event was marked as favorite.

Metawolve:
- Adds a events/statistics endpoint (only for Admins/EventFeedbackManagers), that returns live favorite counts for all events, grouped by user registration status, until the event starts - after event starts, it returns a snapshot of this data from the event start time (approximately)
- Adds registration status to users in database
- Adds job that saves a snapshot of the current favorite counts to the database, once the event starts
- Adds event page with filtering and sorting functionality to Backoffice, with a day filter and favorite counters, a dropdown to select for which user registration status the favorite count should be shown, and an indicator for events which already started, with a hover tooltip that explains that the counter has been locked since event start

image
(old test data)


src/Eurofurence.App.Backoffice/Pages/Events.razor
```diff
@@ -0,0 +1,114 @@
...
+        <MudDataGridPager T="EventResponse" />
+    </PagerContent>
+</MudDataGrid>
+
+@code {
+    private string? _eventSearch;
+    private List<EventResponse> _events = new();
+    private List<EventConferenceDayResponse> _eventConferenceDays = new();
+    private Guid? _selectedEventConferenceDayId;
+    private MudDataGrid<EventResponse>? _dataGrid;
```
maakinoh: Field is assigned but never used

src/Eurofurence.App.Backoffice/Pages/Events.razor
```diff
@@ -0,0 +1,114 @@
+﻿@page "/events"
+@attribute [Authorize(Roles = "Admin")]
+@using Eurofurence.App.Backoffice.Services
+@using Microsoft.AspNetCore.Authorization
+@using Eurofurence.App.Domain.Model.Events
+@inject ISnackbar Snackbar
```
maakinoh: Never used. Can be removed

src/Eurofurence.App.Backoffice/Pages/Events.razor
```diff
@@ -0,0 +1,114 @@
...
+@inject IEventService EventService
+@using MudBlazor;
```
maakinoh: Directive is not required and can be removed

src/Eurofurence.App.Backoffice/Pages/Events.razor
```diff
@@ -0,0 +1,114 @@
...
+    <MudText Typo="Typo.h6">Events</MudText>
+    <MudSpacer />
+    <MudToggleGroup T="Guid?" Color="Color.Primary" Value="_selectedEventConferenceDayId" ValueChanged="EventConferenceDayChanged" Class="mt-1">
+        <MudToggleItem T="Guid?" Value="@(null)">All</MudToggleItem>
+        @foreach (var day in _eventConferenceDays)
+        {
+            <MudToggleItem T="Guid?" Value="@day.Id">@day.Date.ToString("ddd")</MudToggleItem>
+        }
+    </MudToggleGroup>
+    <MudTextField Class="ml-4" T="string" ValueChanged="Search" Clearable="true" Label="Search" Variant="Variant.Outlined" Margin="Margin.Dense"
```
maakinoh, line 19: ValueChanged callback returns a nullable value.
The method at 84-80 should probably be adjusted.

src/Eurofurence.App.Backoffice/Services/EventService.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Net;
```
maakinoh: Not required. Can be removed

src/Eurofurence.App.Backoffice/Services/EventService.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Net.Http.Json;
+using Eurofurence.App.Domain.Model.Events;
+using static System.Net.WebRequestMethods;
```
maakinoh: Not required. Can be removed

maakinoh (changes requested):
Thank you for this PR

I only have some minor feedback.
The rest looks good.

src/Eurofurence.App.Server.Web/Controllers/EventsController.cs
```diff
@@ -153,6 +157,24 @@ public async Task<ActionResult> GetFavoritesUrl()
...
+        /// <returns>
+        /// All events in the event schedule, including statistics on favorite counts.
+        /// </returns>
+        [Authorize(Roles = "Admin,EventFeedbackManager")]
+        [HttpGet("Statistics")]
+        [ProducesResponseType(typeof(string), 404)]
+        [ProducesResponseType(typeof(IEnumerable<EventStatisticsResponse>), 200)]
+        public IEnumerable<EventStatisticsResponse> GetEventStatisticsAsync()
+        {
+            var isStaff = User?.IsInRole("Staff") ?? false;
```
maakinoh: I think `User` is non-nullable since this method has the Authorize attribute.

src/Eurofurence.App.Backoffice/Pages/Events.razor
```diff
@@ -0,0 +1,144 @@
+﻿@page "/events"
+@attribute [Authorize(Roles = "Admin,EventFeedbackManager")]
+@using Eurofurence.App.Backoffice.Services
+@using Microsoft.AspNetCore.Authorization
+@using Eurofurence.App.Domain.Model.Events
+@inject IEventService EventService
+@using MudBlazor;
```
maakinoh: `MudBlazor` import is not needed

```suggestion
```

src/Eurofurence.App.Domain.Model/Events/EventStatisticsResponseRegister.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class EventStatisticsResponseRegister : IRegister
+    {
+        public void Register(TypeAdapterConfig config)
+        {
+            config
+                .NewConfig<EventRecord, EventStatisticsResponse>()
+                .Map(dest => dest.FavoredByCount, src => src.FavoredBy.Count)
+                .Map(dest => dest.FavoredByCheckedInCount, src => src.FavoredBy
+                    .GroupBy(x => x.IdentityId).Select(y => y.FirstOrDefault())
+                    .Count(e => e.RegistrationStatus == UserRegistrationStatus.CheckedIn));
```
maakinoh: Is there a reason why we deduplicate the favorites for check-in users but not the total favorites?
If I remember correctly, a user can't favor an event multiple times.

So the `GroupBy` is probably not needed here - or if there are indeed multiple favs by the same user on the same event, we should deduplicate the events for the total fav count too.

maakinoh (changes requested):
Thanks again for the new changes.

I just have some minor suggestions. The rest looks good.

src/Eurofurence.App.Domain.Model/Events/EventStatisticsResponseRegister.cs
Metawolve: Thanks for your review ^^

With the new changes, we now always groupBy. The total count is now not returned by the API anymore, as we now return the favorite counts for all registration types, and when a user wants the total count he can just add them up (we do this in the Backoffice).

The reason we do the groupBy is, because we are currently not actually storing the users, but instead we are storing registrations from the reg system in our DB. It is theoretically possible for a user to have two registrations in the reg system, which would create two entries with the same identityId in our database. In practice this will probably never happen as we wipe the registrations each year from the IDP afaik, but just to be safe I added the groupBy and firstOrDefault.

We wanted to refactor the user-related stuff, but that's something for next year, as I didn't really get to it this year and... it works for now

src/Eurofurence.App.Backoffice/Services/EventService.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Eurofurence.App.Domain.Model.ArtistsAlley;
```
maakinoh: Import not needed

```suggestion
```

src/Eurofurence.App.Domain.Model/Events/EventFavoriteStatisticsResponse.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Eurofurence.App.Domain.Model.Users;
+using System;
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
```
maakinoh: Import not needed
```suggestion
```

src/Eurofurence.App.Domain.Model/Events/EventWithStatisticsResponse.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Eurofurence.App.Domain.Model.Events;
```
maakinoh: All other classes declare the namespace with curly brackets blocks.
For consistency, consider restyling this. 

Not really a blocking issue though.

src/Eurofurence.App.Domain.Model/Events/EventFavoriteStatisticsResponse.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Eurofurence.App.Domain.Model.Events
+{
+    [DataContract]
+    public class EventFavoriteStatisticsResponse : ResponseBase
+    {
+        [DataMember]
+        public Guid EventId { get; set; }
+
+        [DataMember]
+        public UserRegistrationStatus UserRegistrationStatus { get; set; } = new UserRegistrationStatus();
```
maakinoh: Assuming this field should be initialized with the default value of `UserRegistrationStatus` (Unknown).

```suggestion
        public UserRegistrationStatus UserRegistrationStatus { get; set; } = UserRegistrationStatus.Unknown;
```
Makes it clearer which value is initialized as default.
I think we can even remove it - might be redundant according to rider

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -23,7 +13,20 @@
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
```
maakinoh: Import not needed

```suggestion
```

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -23,7 +13,20 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Eventing.Reader;
```
maakinoh: Import not needed

```suggestion
```

src/Eurofurence.App.Server.Web/Jobs/UpdateEventFavoriteStatisticsJob.cs
```diff
@@ -0,0 +1,40 @@
...
+            _logger.LogInformation(LogEvents.Import, $"Starting job {context.JobDetail.Key.Name}");
+
+            try
+            {
+                await _eventFavoriteStatisticsService.InsertForAllStartedEvents();
+            }
+            catch (Exception e)
+            {
+                SentrySdk.CaptureException(e);
+                _logger.LogError(LogEvents.Import, $"Job {context.JobDetail.Key.Name} failed with exception: {e.Message} {e.StackTrace}");
```
maakinoh: The logger has a specific method overload for logging exceptions.

Also, using string interpolation is not recommended (csharpsquid:S2629) since they introduce unnecessary overhead.

It is recommended (also by https://www.nuget.org/packages/Microsoft.Extensions.Logging) to use the method overload in this case.

```suggestion
                _logger.LogError(LogEvents.Import, e, "Job {Name} failed with exception", context.JobDetail.Key.Name);
```

src/Eurofurence.App.Server.Web/Jobs/UpdateEventFavoriteStatisticsJob.cs
```diff
@@ -0,0 +1,40 @@
...
+            ILoggerFactory loggerFactory,
+            IEventFavoriteStatisticsService eventFavoriteStatisticsService)
+        {
+            _eventFavoriteStatisticsService = eventFavoriteStatisticsService;
+            _logger = loggerFactory.CreateLogger(GetType());
+        }
+
+        public async Task Execute(IJobExecutionContext context)
+        {
+            _logger.LogInformation(LogEvents.Import, $"Starting job {context.JobDetail.Key.Name}");
```
maakinoh: Same logging issue here

```suggestion
            _logger.LogInformation(LogEvents.Import, "Starting job {Name}", context.JobDetail.Key.Name);
```

maakinoh (changes requested):
I just found a few minor smells.

The rest looks good

maakinoh (approved):
LGTM

PR #442, opened 2026-07-12
feat(events): enable tracking of event slots across schedules and setting all events to internal

Fenrikur:
* fixes issue with event entries not being flagged as internal
* adds a configuration key to make all events internal for pre-production use
* fixes an issue where the `SourceId` of an event would change with every new schedule version caused by Pretalx:
  * allowing submissions to be referenced by more than one slot in the schedule while
  * copying all slots for each new schedule release, leading to
  * `slot.id` having a new value with each release of a new schedule version
  * making it impossible to track a slot's movements across schedule releases.

### DIFF
--- a/src/Eurofurence.App.Server.Services/Events/EventService.cs
+++ b/src/Eurofurence.App.Server.Services/Events/EventService.cs
@@ -224,10 +224,10 @@ namespace Eurofurence.App.Server.Services.Events
                 var slotsPublic = _eventOptions.MarkAllInternal ?
                     new HashSet<PretalxSlot>()
                     :
-                    slots.Where(slot =>
+                    [.. slots.Where(slot =>
                         (!slot.Submission?.Tags.Contains(_eventOptions.InternalTagId) ?? false) &&
                         (!slot.Submission?.Track.Id.Equals(_eventOptions.InternalTrackId) ?? false)
-                    ).ToHashSet();
+                    )];
 
                 var tracks = slots.Where(slot => slot.Submission != null).Select(slot => slot.Submission.Track).ToHashSet();
                 var tracksPublic = slotsPublic.Select(slot => slot.Submission.Track).ToHashSet();
@@ -259,7 +259,7 @@ namespace Eurofurence.App.Server.Services.Events
             }
         }
 
-        private string GenerateUniquePretalxSlotSourceId(IEnumerable<PretalxSlot> importEventEntries, PretalxSlot source)
+        private static string GenerateUniquePretalxSlotSourceId(IEnumerable<PretalxSlot> importEventEntries, PretalxSlot source)
         {
             // Slots without submissions attached are either blockers or breaks, so if they move,
             // we just recreate them for the time being. This will break favorites on blockers and
@@ -269,17 +269,24 @@ namespace Eurofurence.App.Server.Services.Events
                     $"BLOCKR-{((DateTimeOffset?)source.Start)?.ToUnixTimeSeconds()}-{((DateTimeOffset?)source.End)?.ToUnixTimeSeconds()}-{source.Room.Id}"
                     )
                 )
-            ).Substring(0, 6);
+            )[..6];
 
             // Since the same submission may be in multiple 
-            var slotIndex = source.Submission != null ? importEventEntries.Where(entry => entry.Submission?.Code == source.Submission?.Code).OrderBy(entry => entry.Start).Index().SingleOrDefault(indexed => indexed.Item.Id == source.Id, new(0, null)).Index.ToString() : "BLOCKER";
+            var slotIndex = source.Submission == null ?
+                "BLOCKER"
+                :
+                importEventEntries.Where(entry => entry.Submission?.Code == source.Submission?.Code)
+                    .OrderBy(entry => entry.Start)
+                    .Index()
+                    .SingleOrDefault(indexed => indexed.Item.Id == source.Id, new(0, null))
+                    .Index.ToString();
 
             return $"{submissionCode}-{slotIndex}";
         }
 
         private async Task<Tuple<int, List<EventConferenceDayRecord>>> UpdateEventConferenceDaysAsync(
             IList<Tuple<DateTime, string>> importDays,
-            ISet<DateTime> importDaysPublic
+            HashSet<DateTime> importDaysPublic
         )
         {
             var eventConferenceDayRecords = _appDbContext.EventConferenceDays.AsNoTracking();
@@ -302,8 +309,8 @@ namespace Eurofurence.App.Server.Services.Events
 
 
         private async Task<Tuple<int, List<EventConferenceTrackRecord>>> UpdateEventConferenceTracksAsync(
-            ISet<PretalxTrack> importTracks,
-            ISet<PretalxTrack> importTracksPublic
+            HashSet<PretalxTrack> importTracks,
+            HashSet<PretalxTrack> importTracksPublic
         )
         {
             var eventConferenceTrackRecords = _appDbContext.EventConferenceTracks.AsNoTracking();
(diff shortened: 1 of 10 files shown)


src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -213,7 +214,20 @@ public async Task RunImportAsync()
...
-                var slotsPublic = slots.Where(slot => (!slot.Submission?.Tags.Contains(_eventOptions.InternalTagId)) ?? false);
+
+                // Try to ensure we have a stable way of identifying slots across multiple schedule versions
+                foreach (var slot in slots)
+                {
+                    slot.SourceId = GenerateUniquePretalxSlotSourceId(slots, slot);
+                }
+
+                var slotsPublic = _eventOptions.MarkAllInternal ?
+                    new HashSet<PretalxSlot>()
```
Metawolve, line 247: Could use collection expression, then you can remove .ToHashSet:

```
new HashSet<PretalxSlot>()
:
[.. slots.Where(slot =>
    (!slot.Submission?.Tags.Contains(_eventOptions.InternalTagId) ?? false) &&
    (!slot.Submission?.Track.Id.Equals(_eventOptions.InternalTrackId) ?? false)
)];
```

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -319,6 +352,7 @@ ISet<PretalxRoom> importRoomsPublic
 
         private async Task<Tuple<int, List<EventRecord>>> UpdateEventEntriesAsync(
             IEnumerable<PretalxSlot> importEventEntries,
+            ISet<PretalxSlot> importEventEntriesPublic,
```
Metawolve: Prefer using HashSet instead of ISet for performance. No need to use interface here

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -328,12 +362,12 @@ IDictionary<int, PretalxTag> tags
...
             var patch = new PatchDefinition<PretalxSlot, EventRecord>(
-                (source, targets) => targets.SingleOrDefault(target => target.SourceId == source.Id)
+                (source, targets) => targets.SingleOrDefault(target => target.SourceId == source.SourceId)
             );
 
             // Public breaks have no submission; their title can be found in the slot description.
-            patch.Map(source => source.Id, target => target.SourceId)
-                .Map(source => $"{source.Submission?.Code ?? "BLOCKER"}-{source.Id}", target => target.Slug)
+            patch.Map(source => source.SourceId, target => target.SourceId)
+                .Map(source => source.Submission?.Code ?? source.SourceId?.Split('-')[0] ?? "UNKNWN", target => target.Slug)
```
Metawolve, line 399: Technically not from this PR but not sure if we need the null conditional operator ? after Submission. PretalxSubmission.Submission is of type PretalxSubmission (not nullable), so if it could be null we should make it nullable, and if it can not be null we can remove the null conditional operators after source.Submission

src/Eurofurence.App.Server.Services/Events/EventService.cs
Metawolve: Could use collection expression here too when we're at it

```
return new Tuple<int, List<EventRecord>>(modifiedRecords, [.. diff.Where(a => a.Entity.IsDeleted == 0).Select(a => a.Entity)]);

```

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -236,14 +251,32 @@ public async Task RunImportAsync()
...
+        {
+            // Slots without submissions attached are either blockers or breaks, so if they move,
+            // we just recreate them for the time being. This will break favorites on blockers and
+            // breaks, but they are internal and limited use only anyways.
+            var submissionCode = source.Submission?.Code ?? Convert.ToHexString(
+                MD5.HashData(System.Text.Encoding.UTF8.GetBytes(
+                    $"BLOCKR-{((DateTimeOffset?)source.Start)?.ToUnixTimeSeconds()}-{((DateTimeOffset?)source.End)?.ToUnixTimeSeconds()}-{source.Room.Id}"
+                    )
+                )
+            ).Substring(0, 6);
```
Metawolve: Could be simplified to
`[..6]`
instead of
`.Substring(0, 6)`

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -236,14 +251,32 @@ public async Task RunImportAsync()
...
+                    $"Event import for schedule version {pretalxSchedule.Version} finished successfully modifying {eventConferenceTracks.Item1} of {tracks.Count} ({tracksPublic.Count()} public) EventConferenceTrack(s), {eventConferenceRooms.Item1} of {rooms.Count} ({roomsPublic.Count()} public) EventConferenceRoom(s), {eventConferenceDays.Item1} of {days.Count} ({daysPublic.Count()} public) EventConferenceDay(s) and {eventEntries.Item1} of {slots.Count()} ({slotsPublic.Count()} public) Event(s) from previously imported version {lastScheduleVersion ?? "<none>"}.");
             }
             finally
             {
                 Semaphore.Release();
             }
         }
 
+        private string GenerateUniquePretalxSlotSourceId(IEnumerable<PretalxSlot> importEventEntries, PretalxSlot source)
```
Metawolve: Method can be made static

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -236,14 +251,32 @@ public async Task RunImportAsync()
...
+            // breaks, but they are internal and limited use only anyways.
+            var submissionCode = source.Submission?.Code ?? Convert.ToHexString(
+                MD5.HashData(System.Text.Encoding.UTF8.GetBytes(
+                    $"BLOCKR-{((DateTimeOffset?)source.Start)?.ToUnixTimeSeconds()}-{((DateTimeOffset?)source.End)?.ToUnixTimeSeconds()}-{source.Room.Id}"
+                    )
+                )
+            ).Substring(0, 6);
+
+            // Since the same submission may be in multiple 
+            var slotIndex = source.Submission != null ? importEventEntries.Where(entry => entry.Submission?.Code == source.Submission?.Code).OrderBy(entry => entry.Start).Index().SingleOrDefault(indexed => indexed.Item.Id == source.Id, new(0, null)).Index.ToString() : "BLOCKER";
```
Metawolve: Bit long and not very readable Linq statement - maybe use early return for "BLOCKER" path and/or use newlines to make it more readable.

Also, as Submission isn't nullable here, either make it nullable in the PretalxSubmission class, or if it can never be null, we can remove the null conditional operator.
Same as https://github.com/eurofurence/ef-app_backend-dotnet-core/pull/442/changes#r3568995220

Metawolve (commented):
Just some nitpicks, mostly syntax stuff.
Otherwhise good solution ^^
Can't think of a better way to keep our events, when schedule in Pretalx changes

src/Eurofurence.App.Server.Services/Events/EventService.cs
Fenrikur: Even though the Pretalx API claims its not nullable, it actually is. Amending the DTO classes accordingly.

src/Eurofurence.App.Server.Services/Events/EventService.cs
Fenrikur: Assuming this is transferrable to all other instances of `ISet<>` in the `EventService`, will fix all other occurrences within other methods on this class, too.

Metawolve (dismissed):
LGTM

PR #443, opened 2026-07-12
feat(sync): AppConfig section for arbitrary key value pairs

Fenrikur:
Allows us to set feature flags and some other configuration values via the API instead of compiling them into the app.

### DIFF
--- a/src/Eurofurence.App.Domain.Model/Storage/AggregatedDeltaResponse.cs
+++ b/src/Eurofurence.App.Domain.Model/Storage/AggregatedDeltaResponse.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using Eurofurence.App.Domain.Model.Announcements;
 using Eurofurence.App.Domain.Model.ArtistsAlley;
 using Eurofurence.App.Domain.Model.Dealers;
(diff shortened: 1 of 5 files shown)


src/Eurofurence.App.Domain.Model/Storage/AggregatedDeltaResponse.cs
```diff
@@ -28,6 +29,7 @@ public class AggregatedDeltaResponse
         public DeltaResponse<AnnouncementResponse> Announcements { get; set; }
         public DeltaResponse<MapResponse> Maps { get; set; }
         public DeltaResponse<ArtistAlleyResponse> TableRegistrations { get; set; }
+        public Dictionary<string, string> AppConfig { get; set; }
```
Metawolve, line 32: I'm fine with putting this in the appsettings.json, pragmatic solution over a database table, assuming these values don't change often and don't need to be changable at runtime (changing appsettings.json at runtime would be possible with some implementation work though, but we currently don't support it).

But if we can, we should really use a real object instead of dynamic key-value pairs.
It would give us API documentation and compile-time safety. Do we really add properties so often or during the con so we really need it to be dynamic?

Metawolve (commented):
Just one comment. Thanks for the PR!

src/Eurofurence.App.Domain.Model/Storage/AggregatedDeltaResponse.cs
Fenrikur: Making these static via a type would very tightly couple backend and frontend without a real need for it. We can potentially consider making some values that will _always_ be needed part of a type, but I would very strongly vote for having a dynamic part since we will be controlling things in the frontend that do not matter to the backend implementation at all.

Values are changed very infrequently, so a brief restart of the service to pull the changed values is not an issue.

Thank you for the review! (:

Metawolve (approved):
LGTM

PR #449, opened 2026-07-14
fix: remove internal events filter from favorite events retrieval

maakinoh:
Internal events are now also included in the calendar export.

We already check if a user is staff when they try to fave an event.


Fenrikur (approved):
Thanks! LGTM :D

PR #455, opened 2026-07-19
feat(pass): support pkpass format for mobile wallets

Fenrikur:


src/Eurofurence.App.Server.Services/Abstractions/Passes/IPassService.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// MIME type definition for Apple's pkpass format.
+        /// </summary>
+        public const string MimeTypePkpass = "application/vnd.apple.pkpass";
+
+        /// <summary>
+        /// Generates a data matrix code for the user based on their registration as SVG.
+        /// </summary>
+        /// <param name="identity">ClaimsIdentity for the registration ID of which the code should be generated.</param>
+        /// <returns>Pass containing SVG image data and metadata or <c>null</c> if no registration is attached to given identity.</returns>
+        public PassFile GenerateDataMatrixCode(ClaimsIdentity identity);
```
Metawolve: If null can be returned, the return type should be PassFile?

You'll need to specify `#nullable enable` on top of the file.
We should probably set `<Nullable>enable</Nullable>` for our solution so we won't have to do that anymore, but that is probably out of scope for this PR I guess

src/Eurofurence.App.Server.Services/Abstractions/Passes/IPassService.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Generates a wallet pass conformant to Apple's pkpass format, which can be imported into
+        /// e.g. Apple or Google wallet.
+        /// </summary>
+        /// <param name="identity">ClaimsIdentity the associated registration of which should be used for generating the pass.</param>
+        /// <returns>Wallet pass containing the registration data or <c>null</c> if no registration is attached to given identity.</returns>
+        public Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, CancellationToken cancellationToken = default);
```
Metawolve, line 32: Same as https://github.com/eurofurence/ef-app_backend-dotnet-core/pull/455/changes#r3625776833 - return type should be `Task<Passfile?>`

src/Eurofurence.App.Server.Services/Abstractions/Passes/PassFile.cs
```diff
@@ -0,0 +1,9 @@
+namespace Eurofurence.App.Server.Services.Abstractions.Passes
+{
+    public class PassFile
+    {
+        public byte[] data { get; init; }
```
Metawolve: These public properties must use pascal casing, according to C# conventions

src/Eurofurence.App.Server.Services/Images/ImageService.cs
```diff
@@ -211,18 +211,19 @@ await UploadFileToMinIoAsync(
             return existingRecord;
         }
 
-        public async Task<Stream> GetImageContentByImageIdAsync(Guid id, CancellationToken cancellationToken = default)
+        public async Task<MemoryStream> GetImageStreamByImageIdAsync(Guid id, CancellationToken cancellationToken = default)
```
Metawolve, line 214: Return type should be `Task<MemoryStream?>` , as null can be returned

src/Eurofurence.App.Server.Services/Images/ImageService.cs
```diff
@@ -211,18 +211,19 @@ await UploadFileToMinIoAsync(
...
         {
-            var existingRecord = await
-                _appDbContext.Images
-                    .AsNoTracking()
-                    .FirstOrDefaultAsync(entity => entity.Id == id, cancellationToken);
+            if (await _appDbContext.Images.AsNoTracking()
```
Metawolve, line 216: This would be a bit nicer to read in my opinion:

```
var existingRecord = await _appDbContext.Images.AsNoTracking()
    .FirstOrDefaultAsync(entity => entity.Id == id, cancellationToken);

if (existingRecord is null)
{
    return null;
}
```

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+        private readonly GlobalOptions _globalOptions;
+        private readonly PassOptions _passOptions;
+        private readonly IHttpClientFactory _httpClientFactory;
+        private readonly HybridCache _cache;
+        private readonly X509Certificate2 _appleWwdrCertificate;
+        private readonly X509Certificate2 _passbookCertificate;
+        private readonly PassGenerator _passGenerator;
+        private readonly ILogger _logger;
+
+        private static string AvatarHttpClientName = "Eurofurence.App.Server.Services.Passes.AvatarHttpClientName";
```
Metawolve: Why use a name for the http client? It can just be created using `_httpClientFactory.CreateClient()` without a name, as it is created inline and not registered in the DI pipeline.

But if we do, prefer `private const`

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+                return null;
+
+            var serialNumber = Convert.ToHexString(
+                MD5.HashData(Encoding.UTF8.GetBytes(
+                    $"{_globalOptions.ConventionIdentifier}-{registrationId}"
+                    )
+                )
+            );
+
+            PassGeneratorRequest request = new PassGeneratorRequest
```
Metawolve: Could be
`var request = new PassGeneratorRequest`
or
`PassGeneratorRequest request = new()`

Currently causes IDE language flag in VS

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Logo2X", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.Logo2XImageId, cancel), cancellationToken: cancellationToken)
+            );
+            request.Images.Add(PassbookImage.Logo3X,
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Logo3X", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.Logo3XImageId, cancel), cancellationToken: cancellationToken)
+            );
+            request.Images.Add(PassbookImage.Background,
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Background", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.BackgroundImageId, cancel), cancellationToken: cancellationToken)
+            );
+
+            if (identity.Claims.FirstOrDefault(c => c.Type == "avatar")?.Value is var avatarUrl && avatarUrl.StartsWith("https://"))
```
Metawolve: Should check for `string.IsNullOrEmpty`. Currently, `avatarUrl ` could be null

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
+using Eurofurence.App.Server.Services.Abstractions.Identity;
+using Microsoft.Extensions.Caching.Distributed;
```
Metawolve: Unused import

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+using System;
+using System.Security.Cryptography;
+using Eurofurence.App.Server.Services.Abstractions.Images;
+using Microsoft.Extensions.Caching.Hybrid;
+using System.Threading.Tasks;
+using System.Threading;
+using Microsoft.Extensions.Logging;
+using Passbook.Generator.Fields;
+using System.Text.Json.Nodes;
+using System.IO;
```
Metawolve: More unused imports

src/Eurofurence.App.Server.Services/Abstractions/Passes/IPassService.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using System.IO;
```
Metawolve: Unused imports

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+        {
+            _identityService = identityService;
+            _imageService = imageService;
+            _globalOptions = globalOptions.Value;
+            _passOptions = passOptions.Value;
+            _httpClientFactory = httpClientFactory;
+            _cache = cache;
+            _passGenerator = new PassGenerator();
+            _appleWwdrCertificate = X509Certificate2.CreateFromPem(_passOptions.AppleWwdrX509CertificatePem);
+            _passbookCertificate = X509Certificate2.CreateFromPem(_passOptions.PassbookX509CertificatePem, _passOptions.PassbookX509KeyPem);
```
Metawolve: This service is added with `.AddTransient` (like most of our services) which means, that these certificates will be parsed on every single API request, which is expensive. We could create a provider service using `.AddSingleton`, for example `IPassCertificateProvider`, that loads the certs once on startup and inject that service here.

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+            request.Images.Add(PassbookImage.Background,
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Background", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.BackgroundImageId, cancel), cancellationToken: cancellationToken)
+            );
+
+            if (identity.Claims.FirstOrDefault(c => c.Type == "avatar")?.Value is var avatarUrl && avatarUrl.StartsWith("https://"))
+            {
+                try
+                {
+                    using var httpClient = _httpClientFactory.CreateClient(AvatarHttpClientName);
+                    var avatarImage = await httpClient.GetByteArrayAsync(avatarUrl);
```
Metawolve: Missing passing of `cancellationToken`

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        public PassFile GenerateDataMatrixCode(ClaimsIdentity identity)
+        {
+            var registrationId = _identityService.GetRegistrationsIds(identity).FirstOrDefault();
+
+            if (string.IsNullOrEmpty(registrationId))
+                return null;
+
+            BarcodeWriterSvg writer = new() { Format = BarcodeFormat.AZTEC };
```
Metawolve, line 65: Method name is `GenerateDataMatrixCode`, but it generates a Aztec code

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        public async Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, CancellationToken cancellationToken = default)
+        {
+            var registrationId = _identityService.GetRegistrationsIds(identity).FirstOrDefault();
+
+            if (string.IsNullOrEmpty(registrationId))
+                return null;
+
+            var serialNumber = Convert.ToHexString(
+                MD5.HashData(Encoding.UTF8.GetBytes(
```
Metawolve: Even if this is not a security concern, using MD5 can cause security warnings in some tools, prefer using `SHA256.HashData`, it's not too much slower

src/Eurofurence.App.Server.Services/Passes/PassService.cs
Metawolve: Could add timeout to the http client to prevent thread blocking

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Logo", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.LogoImageId, cancel), cancellationToken: cancellationToken)
+            );
+            request.Images.Add(PassbookImage.Logo2X,
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Logo2X", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.Logo2XImageId, cancel), cancellationToken: cancellationToken)
+            );
+            request.Images.Add(PassbookImage.Logo3X,
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Logo3X", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.Logo3XImageId, cancel), cancellationToken: cancellationToken)
+            );
+            request.Images.Add(PassbookImage.Background,
+                await _cache.GetOrCreateAsync("PassService.Pkpass.Background", async cancel => await _imageService.GetImageContentByImageIdAsync(_passOptions.BackgroundImageId, cancel), cancellationToken: cancellationToken)
```
Metawolve, line 243: Caching the images as byte[] in the cache will hog memory and is bad practice. The cache is for small amounts of data, not whole files.
We should use Streaming or at least just cache the URLs of the images, not the images themselves..

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+            request.SemanticTags.Add(new VenueRegionName(_globalOptions.ConventionVenueRegion));
+            request.SemanticTags.Add(new VenueRoom(_passOptions.VenueRoom));
+            request.SemanticTags.Add(new VenueLocation(
+                _globalOptions.ConventionVenueLocationLatitude,
+                _globalOptions.ConventionVenueLocationLongitude
+                )
+            );
+
+            request.SemanticTags.Add(new AttendeeName(identity.Name));
+            request.SemanticTags.Add(new AdmissionLevel("Cookie Connoisseur"));
```
Metawolve: Is this used? :D 
Could be a constant at least

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+            _passbookCertificate = X509Certificate2.CreateFromPem(_passOptions.PassbookX509CertificatePem, _passOptions.PassbookX509KeyPem);
+            _logger = loggerFactory.CreateLogger(GetType());
+        }
+
+        public PassFile GenerateDataMatrixCode(ClaimsIdentity identity)
+        {
+            var registrationId = _identityService.GetRegistrationsIds(identity).FirstOrDefault();
+
+            if (string.IsNullOrEmpty(registrationId))
+                return null;
```
Metawolve, line 63: Should we add logging here?

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -0,0 +1,296 @@
...
+        /// Name of the city (e.g. "Hamburg, Germany").
+        /// </param>
+        private class VenueRegionName(string value) : SemanticTagBaseValue("venueRegionName", value);
+
+        public async Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, CancellationToken cancellationToken = default)
+        {
+            var registrationId = _identityService.GetRegistrationsIds(identity).FirstOrDefault();
+
+            if (string.IsNullOrEmpty(registrationId))
+                return null;
```
Metawolve, line 111: Should we add logging here?

Metawolve (commented):
Good PR overall, makes sense to add a service for the Passes.

Some suggestions - especially caching the images as byte[] in HybridCache and parsing the certs in PassService's constructor is something we should overthink. The rest is mostly syntaxy stuff

src/Eurofurence.App.Server.Services/Abstractions/Passes/IPassService.cs
Fenrikur: Separate task; fix on global scale and not per file.

src/Eurofurence.App.Server.Services/Images/ImageService.cs
Fenrikur: Separate task; fix on global scale and not per file.

src/Eurofurence.App.Server.Services/Passes/PassService.cs
Fenrikur: The idea behind this is to avoid all HTTP requests associated with loading all of the images from minIO with every request. If this consumes a few MB of RAM at runtime, I would see this as the lesser evil?

src/Eurofurence.App.Server.Services/Passes/PassService.cs
Fenrikur: Why? It just means that somebody who does not have a registration tried to request a pass, which cannot be issued. Doesn't seem notable enough to produce log output for.

src/Eurofurence.App.Server.Services/Passes/PassService.cs
Fenrikur: Why? It just means that somebody who does not have a registration tried to request a pass, which cannot be issued. Doesn't seem notable enough to produce log output for.

Metawolve (dismissed):
LGTM now 👍 thank you

Metawolve (approved):
LGTM!

PR #456, opened 2026-07-21
fix(event): use UTC timezone for calendar export

maakinoh:
We need to explicitly set the time to UTC when adding events to the calendar.


Fenrikur (approved):
lgtm, thanks!

PR #457, opened 2026-07-21
fix(event): avoid trying to add same favorite again

Fenrikur:


maakinoh (dismissed):
Thank you for this PR

LGTM

src/Eurofurence.App.Server.Services/Events/EventService.cs
```diff
@@ -83,12 +83,11 @@ public async Task AddEventToFavoritesIfNotExist([NotNull] ClaimsPrincipal user,
                 .Include(userRecord => userRecord.FavoriteEvents)
                 .FirstOrDefault(x => x.IdentityId == user.GetSubject());
 
-            if (userRecord != null && !userRecord.FavoriteEvents.Contains(eventRecord))
+            if (userRecord != null && userRecord.FavoriteEvents.FirstOrDefault(favorite => favorite.Id == eventRecord.Id) is null)
```
Metawolve: Could use .Any(), bit simpler and more performant, because it stops looking after the first match and doesn't need to return the record:
`if (userRecord != null && !userRecord.FavoriteEvents.Any(favorite => favorite.Id == eventRecord.Id))`

Metawolve (commented):
Just one small suggestion

Metawolve (approved):
LGTM!

PR #484, opened 2026-08-02
fix(identity): avoid caching userinfo without name claim

Fenrikur:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of incomplete authentication information.
  - Malformed authentication now returns a clear error instead of creating a pass token.
  - Invalid user information is excluded from caching to prevent reuse of incomplete data.
  - Added logging to help diagnose authentication data issues.


Metawolve (approved):
LGTM

PR #503, opened 2026-08-08
feat: editing announcements in backoffice

maakinoh:


src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
```diff
@@ -0,0 +1,203 @@
...
+                    <MudItem xl="6" sm="12">
+                        <MudDatePicker Label="End Date" Required="true" @bind-Date="_endDate"/>
+                    </MudItem>
+                    <MudItem xl="6" sm="12">
+                        <MudTimePicker Label="End Time" Required="true" @bind-Time="_endTime"/>
+                    </MudItem>
+                </MudGrid>
+
+                <MudTextField Label="Groups (comma separated)" @bind-Value="_groups"/>
+                <MudTextField Label="Image ID" @bind-Value="Announcement.ImageId"/>
```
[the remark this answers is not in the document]
maakinoh: fixed

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
```diff
@@ -0,0 +1,203 @@
...
+        
+        AnnouncementRequest request = new()
+        {
+            Title = Announcement.Title,
+            Author = Announcement.Author,
+            Area = Announcement.Area,
+            Content = Announcement.Content,
+            ImageId = Announcement.ImageId,
+            ValidFromDateTimeUtc = Announcement.ValidFromDateTimeUtc,
+            ValidUntilDateTimeUtc = endDateTime.Value,
```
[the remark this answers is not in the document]
maakinoh: The input is alreaedy expected to be UTC

Fenrikur (dismissed):
Thank you for the PR; deployed locally and it's coming along nicely!

Some minor functional additions would be good:
* Add `AnnouncementManager` role to allow granting access specifically to this feature (should be able to create/update/delete announcements).
* Allow for announcements to be deleted using the backoffice via delete button next to the edit button, showing a simple modal prompt for verification to prevent accidental deletions.
* Allow changing the `ValidFromDateTimeUtc` field as well in the backoffice.
* (Optional) allow showing soft-deleted announcements for `Admin` and `AnnouncementManager`roles.

src/Eurofurence.App.Domain.Model/Announcements/AnnouncementResponseRegister.cs
```diff
@@ -0,0 +1,13 @@
+using Mapster;
+
+namespace Eurofurence.App.Domain.Model.Announcements
+{
+    public class AnnouncementResponseRegister : IRegister
+    {
+        public void Register(TypeAdapterConfig config)
+        {
+            config.NewConfig<AnnouncementRecord, AnnouncementResponse>()
+                .Map(x => x.Roles, x => x.Groups);
```
Metawolve: Should we instead unify the naming, so we don't need this mapper?

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
```diff
@@ -0,0 +1,266 @@
...
+                        <MudItem xl="6" md="6" xs="12">
+                            <MudDatePicker Label="End Date" Required="true" @bind-Date="_endDate"/>
+                        </MudItem>
+                        <MudItem xl="6" md="6" xs="12">
+                            <MudTimePicker Label="End Time" Required="true" @bind-Time="_endTime"/>
+                        </MudItem>
+                    </MudGrid>
+
+                    <MudTextField Label="Groups (comma separated)" @bind-Value="_groups"/>
+                    <MudTextField Label="Image ID" Validation="@(new Func<string, Task<string>>(ValidateImageId))"
```
Metawolve: Currently, an image ID must be provided. In other places, we have an upload image button. Was it a deliberate decision that an existing image ID must be provided? An upload button would probably be nicer to use, even for Admins

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
```diff
@@ -0,0 +1,266 @@
...
+                        </MudItem>
+                        <MudItem xl="6" md="6" xs="12">
+                            <MudDatePicker Label="End Date" Required="true" @bind-Date="_endDate"/>
+                        </MudItem>
+                        <MudItem xl="6" md="6" xs="12">
+                            <MudTimePicker Label="End Time" Required="true" @bind-Time="_endTime"/>
+                        </MudItem>
+                    </MudGrid>
+
+                    <MudTextField Label="Groups (comma separated)" @bind-Value="_groups"/>
```
Metawolve: Do we even need this textbox here? As I understand it, this is just for editing existing announcements, so after the edit it will not be re-sent as a notification, right? The group IDs here are only for the push notification I think. So it wouldn't do anything to change the IDs here. So maybe this can be removed.

If we want to keep it: I think it could be unclear, what needs to be provided here - maybe at least change the label to what exactly should be entered, for example ("Group IDs from IDP, comma-separated"), also add a hint that if no groups are provided, the announcement will be sent out to all users. Ideally, we would have a multi-select combobox here, with the available groups listed by name, but I'm not sure if we have those available already. If not, probably not worth the extra effort

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
Metawolve: If we want to allow changing the image, we should probably also show the existing image here if the announcement has one. Could probably copy some of the code from ImageDialog to make the UX similar

Metawolve (commented):
Just some comments - mostly on the two textboxes where the user has to provide GUIDs  currently. Apart from that, and code-wise, very good!

src/Eurofurence.App.Domain.Model/Announcements/AnnouncementResponseRegister.cs
maakinoh: Idk if we use that field in the app already.
Don't want to break things there. Otherwise, this would probably be the better approach.

Metawolve: @Fenrikur I don't think sending announcements to specific roles via our Backend is used at all currently, is it?
So it should be fine changing the name of the property?

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
Metawolve: Feedback from Fen: This can be removed

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
Metawolve: Feedback from Fen: If we can copy the code for showing the image and having an upload button from ImageDialog, let's do it - maybe even make it a reusable component (or we do that in another PR) - if it's a lot more effort, keep it as-is for now

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
maakinoh: No, we're not resending the announcements after an edit.
But when we change the groups afterward, the announcements would be visible for the new users in the app.

But I think we don't have access to all available IDP groups, though.
The new label, however, is a good idea.

src/Eurofurence.App.Domain.Model/Announcements/AnnouncementResponseRegister.cs
Fenrikur: We especially should make sure to not mix up terms:
`Roles` => defined and granted by our backend internally, usually based on IDP groups a user has.
`Groups` => whatever IDs the IDP ships

afaik we currently technically only support sending announcements to `Groups` and that only in a limited capacity for groups that we have cached, since we cannot access that information directly on the IDP. It is currently a feature that is not used anywhere though. Should definitely only be `Groups` though since sending announcement to `Roles` has no real use-case. :)

maakinoh: Field was renamed

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
maakinoh: Moved image selection into its own component.

We need to change other components (like ImageDialog) too then.
This is probably something for a new PR, though.

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
maakinoh: Field was removed

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
```diff
@@ -0,0 +1,313 @@
...
+@inject ISnackbar Snackbar
+
+<MudPaper Class="pa-4">
+    <MudDialog>
+        <DialogContent>
+            @if (Announcement != null)
+            {
+                <MudGrid>
+                    <MudItem xs="12" md="5">
+                        <ImageField ImageResponse="@_image" ImageChanged="OnImageUploaded"/>
```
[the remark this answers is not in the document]
maakinoh: fixed

src/Eurofurence.App.Backoffice/Components/EditAnnouncementDialog.razor
```diff
@@ -0,0 +1,313 @@
...
+        {
+            await FetchImage(imageId);
+        }
+    }
+
+    private async Task FetchImage(Guid imageId)
+    {
+        ImageResponse? image = await ImageService.GetImageAsync(imageId);
+        _image = image;
+        StateHasChanged();
```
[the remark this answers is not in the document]
maakinoh: fixed

src/Eurofurence.App.Backoffice/Pages/Announcements.razor
```diff
@@ -0,0 +1,135 @@
...
+        IDialogReference dialogRef = await DialogService.ShowAsync<ConfirmDialog>("Confirm", dialogUserban, optionsUserBan);
+        DialogResult? result = await dialogRef.Result;
+
+        if (result is { Canceled: false })
+        {
+            await AnnouncementService.DeleteAnnouncementAsync(contextItem.Id);
+            Snackbar.Add("Announcement deleted", Severity.Success);
+            await LoadAnnouncementsAsync();
+            StateHasChanged();
+        }
```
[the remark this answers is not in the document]
maakinoh: fixed

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -73,7 +73,7 @@ public async Task<AnnouncementResponse> GetAnnouncementAsync([FromRoute] Guid id
         /// <param name="id">ID of the announcement to be deleted</param>
         /// <returns></returns>
         [HttpDelete("{id}")]
-        [Authorize(Roles = IdentityRoles.Admin)]
+        [Authorize(Roles = IdentityRoles.AnnouncementManager)]
         [ProducesResponseType(204)]
         [ProducesResponseType(typeof(string), 404)]
         public async Task<ActionResult> DeleteAnnouncementAsync([FromRoute] Guid id)
```
Fenrikur, line 79: Please allow both `Admin` and `AnnouncementManager` here. (=

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -93,7 +93,7 @@ public async Task<ActionResult> DeleteAnnouncementAsync([FromRoute] Guid id)
         /// <param name="request">New announcement to be pushed</param>
         /// <returns></returns>
         [HttpPost]
-        [Authorize(Roles = IdentityRoles.Admin)]
+        [Authorize(Roles = IdentityRoles.AnnouncementManager)]
         [ProducesResponseType(typeof(Guid), 200)]
         [ProducesResponseType(typeof(string), 409)]
         public async Task<ActionResult> PostAnnouncementAsync([EnsureNotNull][FromBody] AnnouncementRequest request)
```
Fenrikur, line 99: Please allow both `Admin` and `AnnouncementManager` here. (=

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -126,7 +126,7 @@ public async Task<ActionResult> PostAnnouncementAsync([EnsureNotNull][FromBody]
         /// <param name="request">Updated announcement record</param>
         /// <returns></returns>
         [HttpPut("{id}")]
-        [Authorize(Roles = IdentityRoles.Admin)]
+        [Authorize(Roles = IdentityRoles.AnnouncementManager)]
         [ProducesResponseType(204)]
         [ProducesResponseType(typeof(string), 404)]
         public async Task<ActionResult> PutAnnouncementAsync([FromRoute] Guid id,
```
Fenrikur, line 132: Please allow both `Admin` and `AnnouncementManager` here. (=

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -156,7 +156,7 @@ public async Task<ActionResult> PutAnnouncementAsync([FromRoute] Guid id,
         /// </summary>
         /// <returns></returns>
         [HttpDelete]
-        [Authorize(Roles = IdentityRoles.Admin)]
+        [Authorize(Roles = IdentityRoles.AnnouncementManager)]
         [ProducesResponseType(204)]
         public async Task<ActionResult> ClearAnnouncementAsync()
```
Fenrikur, line 161: Please allow both `Admin` and `AnnouncementManager` here. (=

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
```diff
@@ -52,7 +53,7 @@ public override IQueryable<AnnouncementRecord> FindAll()
             }
 
             return _appDbContext.Announcements
-                .Where(entity => entity.Groups == null || !entity.Groups.Any() || entity.Groups.Any(group => userGroups.Contains(group)))
+                .Where(entity => entity.Groups == null || !entity.Groups.Any() || entity.Groups.Any(group => userGroups.Contains(group) || _httpContext.User.IsInRole(IdentityRoles.AnnouncementManager)))
                 .AsNoTracking();
         }
```
Fenrikur: `AnnouncementManager`s should not simply receive all announcements by default, but instead (together with `Admin`s) be able to use an additional `:all` endpoint for `GET` requests, similar to `GetAllImagesAsync()` on [`ImagesController`](https://github.com/eurofurence/ef-app_backend-dotnet-core/blob/main/src/Eurofurence.App.Server.Web/Controllers/ImagesController.cs#L51)
This role should only be relevant in the backoffice, whereas the current implementation would make it also affect the users' frontend/app experience.

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
```diff
@@ -82,7 +83,7 @@ public override async Task<DeltaResponse<AnnouncementResponse>> GetDeltaResponse
                     _appDbContext.Announcements
                         .Where(entity =>
                             entity.IsDeleted == 0
-                            && (entity.Groups == null || !entity.Groups.Any() || entity.Groups.Any(group => userGroups.Contains(group))))
+                            && (entity.Groups == null || !entity.Groups.Any() || entity.Groups.Any(group => userGroups.Contains(group)) || _httpContext.User.IsInRole(IdentityRoles.AnnouncementManager)))
                         .Select(x => x.Transform()).ToArrayAsync(cancellationToken);
```
Fenrikur, line 86: Please remove `AnnouncementManager` from this line as they will otherwise also see all announcements in the app, which is not the intention behind this role. It is purely focused on the backoffice. (=

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
```diff
@@ -93,7 +94,7 @@ public override async Task<DeltaResponse<AnnouncementResponse>> GetDeltaResponse
                     .IgnoreQueryFilters()
                     .Where(entity =>
                         entity.LastChangeDateTimeUtc > minLastDateTimeChangedUtc
-                        && (entity.Groups == null || !entity.Groups.Any() || entity.Groups.Any(group => userGroups.Contains(group))));
+                        && (entity.Groups == null || !entity.Groups.Any() || entity.Groups.Any(group => userGroups.Contains(group)) || _httpContext.User.IsInRole(IdentityRoles.AnnouncementManager)));
```
Fenrikur, line 97: Please remove `AnnouncementManager` from this line as they will otherwise also see all announcements in the app, which is not the intention behind this role. It is purely focused on the backoffice. (=

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
maakinoh: fixed

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
maakinoh: fixed

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
maakinoh: fixed

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
maakinoh: fixed

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
maakinoh: fixed

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
maakinoh: I added a new endpoint, which will be used by the backoffice.

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
maakinoh: fixed

src/Eurofurence.App.Server.Services/Abstractions/Announcements/IAnnouncementService.cs
```diff
@@ -1,10 +1,27 @@
...
 {
     public interface IAnnouncementService :
         IEntityServiceOperations<AnnouncementRecord, AnnouncementResponse>,
         IPatchOperationProcessor<AnnouncementRecord>
     {
+        /// <summary>
+        /// Fetches all announcements regardless of whether the user is a member of the group or not.
+        /// </summary>
+        /// <returns>Collection of all found records.</returns>
+        public IQueryable<AnnouncementRecord> FetchAll();
```
Metawolve: We should think of better naming - difference between FindAll and FetchAll ist unclear. Maybe FindAllForAllGroups or something like that?

src/Eurofurence.App.Server.Services/Abstractions/Announcements/IAnnouncementService.cs
```diff
@@ -1,10 +1,27 @@
...
+
+        /// <summary>
+        /// Finds an announcement by its id regardless of whether the user is a member of the group or not.
+        /// </summary>
+        /// <param name="id">The id to look up.</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>A task with the record. Can be null.</returns>
+        public Task<AnnouncementRecord> FindOneInAllAsync(Guid id, CancellationToken cancellationToken = default);
```
Metawolve: Is this not also just a duplicate of the EntityServiceBase's FindOneAsync method?

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
```diff
@@ -13,6 +13,7 @@
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
+using Eurofurence.App.Domain.Model.Identity;
```
Metawolve: Unused using?

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
```diff
@@ -36,12 +37,26 @@ IIdentityService identityService
...
+        }
+
+        /// <inheritdoc />
+        public async Task<AnnouncementRecord> FindOneInAllAsync(Guid id, CancellationToken cancellationToken = default)
+        {
+            return await _appDbContext.Announcements.AsNoTracking().FirstOrDefaultAsync(entity => entity.Id == id, cancellationToken);
+        }
+
+        /// <inheritdoc />
         public override async Task<AnnouncementRecord> FindOneAsync(Guid id, CancellationToken cancellationToken = default)
```
Metawolve: Not part of your PR but I'm not sure why this is overriden here, it just does the same as the base classes method

src/Eurofurence.App.Server.Services/Announcements/AnnouncementService.cs
```diff
@@ -36,12 +37,26 @@ IIdentityService identityService
...
         }
 
+        /// <inheritdoc />
+        public IQueryable<AnnouncementRecord> FetchAll()
+        {
+            return _appDbContext.Announcements.AsNoTracking();
+        }
+
+        /// <inheritdoc />
+        public async Task<AnnouncementRecord> FindOneInAllAsync(Guid id, CancellationToken cancellationToken = default)
```
Metawolve: Seems to do the same as the FindOneAsync method here, as that one also doesn't check the group

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
Metawolve, line 54: This method should probably just be called GetAnnouncements, instead of GetAnnouncementEntries

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -151,12 +162,35 @@ public async Task<ActionResult> PutAnnouncementAsync([FromRoute] Guid id,
...
         }
 
+        /// <summary>
+        /// Returns the announcement record for the given id.
+        /// If the user is an admin or announcement manager, the record is returned regardless of whether
+        /// they are a member of the groups or not.
+        /// </summary>
+        /// <param name="id">The id to look up.</param>
+        /// <returns>Task with the record. Can be null.</returns>
+        private async Task<AnnouncementRecord> QueryRecordForIdAsync(Guid id)
```
Metawolve: Not sure about the naming here either. Name doesn't really say what the difference to FindOneAsync is. Maybe just put the filtering by role into the FindOneAsync method override and remove this one

src/Eurofurence.App.Server.Web/Controllers/AnnouncementsController.cs
```diff
@@ -44,6 +52,21 @@ public AnnouncementsController(IAnnouncementService announcementService, IImageS
...
         [ProducesResponseType(typeof(IEnumerable<AnnouncementResponse>), 200)]
         public IEnumerable<AnnouncementResponse> GetAnnouncementEntries()
+        {
+            IEnumerable<string> userGroups = new List<string>();
+            if (User?.Identity is ClaimsIdentity identity)
+            {
+                userGroups = _identityService.GetUserGroups(identity);
+            }
+
+            return _announcementService.FindAll(announcement => announcement.Groups == null || !announcement.Groups.Any() || announcement.Groups.Any(group => userGroups.Contains(group))).Select(x => x.Transform());
```
Metawolve, line 62: This should go to the service layer - Controllers should hold no business logic if possible, and also having it in the service makes sure that we have this filtering everywhere we query announcements

Metawolve: After talking to Fen: We will do it like this for now, as we do this in the Controller in other places aswell. Will probably refactor that in general after EF30

Metawolve (dismissed):
LGTM - will create a separate issue for refactoring the permission group checks/entity filtering a bit

Metawolve (approved):
LGTM

PR #505, opened 2026-08-08
fix(pass): perform basic config validation

Fenrikur:


src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -133,11 +147,36 @@ public async Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, Cancell
...
 
-            request.RelevantLocations.Add(new RelevantLocation
+            request.SemanticTags.Add(new EventName($"{_globalOptions.ConventionName} {_globalOptions.ConventionNumber}"));
+            request.SemanticTags.Add(new EventType(EventTypes.PKEventTypeConvention));
+            if (_globalOptions.ConventionVenueName is { Length: > 0 } conventionVenueName)
+                request.SemanticTags.Add(new VenueName(conventionVenueName));
+            if (_globalOptions.ConventionVenueRegion is { Length: > 0 } conventionVenueRegion)
+                request.SemanticTags.Add(new VenueRegionName(conventionVenueRegion));
+            if (_passOptions.VenueRoom is { Length: > 0 } venueRoom)
+                request.SemanticTags.Add(new VenueRoom(venueRoom));
```
Metawolve: Would probably be nicer to use string.IsNullOrWhiteSpace for checks like this. Checks for null, empty, or only-whitespace characters

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -133,11 +147,36 @@ public async Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, Cancell
...
+
+            // TODO: Get Admission Level (attendee/staff/director & regular/sponsor/super sponsor)
+            // from Claims and Registration System
+            //request.SemanticTags.Add(new AdmissionLevel("Cookie Connoisseur"));
+
+            if (_globalOptions.ConventionVenueLocationLatitude is { } conventionVenueLocationLatitude &&
+                    _globalOptions.ConventionVenueLocationLongitude is { } conventionVenueLocationLongitude)
```
Metawolve, line 163: Not sure if these checks do anything currently. ConventionVenueLocationLatitude and ConventionVenueLocationLongitude are not nullable, so they default to 0.0. Would probably be better to check if they are not 0.0, and no need to copy them to a local variable then

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -170,78 +209,101 @@ public async Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, Cancell
...
                 TextAlignment = FieldTextAlignment.PKTextAlignmentRight
             });
+            request.SemanticTags.Add(new EventEndDate(_globalOptions.ConventionEndDateTime.ToString("o", System.Globalization.CultureInfo.InvariantCulture)));
 
-            request.AddBackField(new StandardField
+            /*
+             * Add additional information to back of pass
+             */
+
+            if (identity.Name is { Length: > 0 } attendeeName)
```
Metawolve: Also prefer string.IsNullOrWhiteSpace, also no need to copy to a local variable

src/Eurofurence.App.Server.Services/Passes/PassService.cs
```diff
@@ -170,78 +209,101 @@ public async Task<PassFile> GeneratePkpassAsync(ClaimsIdentity identity, Cancell
...
+            if (_globalOptions.ConventionContact is { Length: > 0 } conventionContact)
+                request.AddBackField(new StandardField
+                {
+                    Key = "contact",
+                    Label = "Contact",
+                    Value = conventionContact,
+                    TextAlignment = FieldTextAlignment.PKTextAlignmentLeft
+                });
+
+            if (_passOptions.Information is { Length: > 0 } passInformation)
```
Metawolve: Also prefer string.IsNullOrWhiteSpace and not copying to a local variable for these checks

Metawolve (commented):
Just some nitpicks on the validation checks, otherwhise good^^

Metawolve (approved):
LGTM

PR #506, opened 2026-08-09
fix: sort announcements backoffice

maakinoh:
Sort announcements by ValidFromDateTimeUtc by default.

Also adding LastChangeDateTimeUtc to the table.


Fenrikur (approved):
LGTM

Thanks!

PR #517, opened 2026-08-13
feat(announcements): set validity via configuration

Fenrikur:


src/Eurofurence.App.Server.Services/Abstractions/Announcements/AnnouncementOptions.cs
```diff
@@ -1,7 +1,16 @@
...
-        public string Url { get; init; }
+        /// <summary>
+        /// URL of the endpoint providing the latest announcements.
+        /// </summary>
+        public string? Url { get; init; }
+        /// <summary>
+        /// Hours from creation that an announcement should be considered valid.
+        /// Defaults to half a year.
+        /// </summary>
+        public int AnnouncementValidityHours { get; init; } = 24 * 183;
```
Metawolve, line 14: As this is already in the announcement section, I think calling this just ValidityHours or ExpirationHours would be a bit simpler

src/Eurofurence.App.Server.Web/appsettings.sample.json
```diff
@@ -154,7 +154,8 @@
     "Password": ""
   },
   "Announcements": {
-    "Url": ""
+    "Url": "",
+    "AnnouncementValidityHours": 4368
```
Metawolve, line 158: Would need to rename here too:)

Metawolve (commented):
Just a naming suggestion, but looks good and happy we have it configurable for this year!

src/Eurofurence.App.Server.Services/Abstractions/Announcements/AnnouncementOptions.cs
Fenrikur: This will be the last year we're using this route, so I'll just keep the slightly more verbose name for now.

src/Eurofurence.App.Server.Web/appsettings.sample.json
Fenrikur: This will be the last year we're using this route, so I'll just keep the slightly more verbose name for now.

PR #520, opened 2026-08-15
fix(announcements): only delete expired entries

Fenrikur:


src/Eurofurence.App.Server.Web/Jobs/UpdateAnnouncementsJob.cs
```diff
@@ -126,7 +126,15 @@ public async Task Execute(IJobExecutionContext context)
...
-                    .Where(a => !string.IsNullOrEmpty(a.Entity.ExternalReference) && a.Action != ActionEnum.NotModified)
+                    .Where(a =>
+                        !string.IsNullOrEmpty(a.Entity.ExternalReference) &&
+                        a.Action != ActionEnum.NotModified &&
+                        (
+                            // Only delete announcements that are no longer valid.
+                            a.Action == ActionEnum.Delete &&
+                            a.Entity.ValidUntilDateTimeUtc.CompareTo(DateTime.UtcNow) < 0
+                        )
+                    )
```
Metawolve, line 140: This is all && linked, so this would mean that only announcements with delete action and that are not valid land in the diff. Added and modified announcements would be ignored, even with the brackets.

I think the correct logic would be:

```
.Where(a =>
    !string.IsNullOrEmpty(a.Entity.ExternalReference) &&
    a.Action != ActionEnum.NotModified &&
    (a.Action != ActionEnum.Delete || a.Entity.ValidUntilDateTimeUtc.CompareTo(DateTime.UtcNow) < 0)
)
```

Metawolve (commented):
I think I found one critical bug in the PR. Otherwhise it looks good!

src/Eurofurence.App.Server.Web/Jobs/UpdateAnnouncementsJob.cs
Fenrikur: Good catch, thank you! Should be fixed now. ♥ 
(And I added some comments for good measure, simply to explain it to myself as I was validating it.)

Metawolve (approved):
LGTM now!

